PR #4, opened 2019-02-20
Optimize memory usage.

yunhailuo:
In general, I pass file path between functions rather and pandas DataFrame so that there won't be a lot of dfs stay in the memory for more time than needed.

### DIFF
--- a/Resources/make_metadata.py
+++ b/Resources/make_metadata.py
@@ -1,0 +1,131 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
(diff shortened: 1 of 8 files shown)


yunhailuo:
These are codes facilitating:
1. ETL large dataset
2. Compare current Xena and GDC datasets, find differences so that only extra datasets need to be loaded

I definitely used them before. But I never pushed them. I'll merge them after some code reviews. @maryjgoldman

yunhailuo:
I think the code is fine. Will merge after #6 .

yunhailuo:
Merge all these before doing structural changes

Issue #1, opened 2019-02-19
Add error check in get_ext(file_name) function in gdc.py


arp95:
Hi,

The get_ext(filename) functions splits the filename on '.' and checks for extensions supported by a constant called _SUPPORTED_FILE_TYPES. 

There is a case in which if the filename doesnt have any supported format available in the _SUPPORTED_FILE_TYPES array, in that case we are returning the format of the file which is not supported by our code. We should add an error check and return NULL or a default extension supported by _SUPPORTED_FILE_TYPES array incase the filename doesnt have a valid ext supported by our code.

maryjgoldman:
@jingchunzhu

yunhailuo:
Thanks @arp95. I'll look into it and see which one of None and default is better.

Ayoob7:
Hello, I would like to fix the issue with the function, whats the procedure to push my branch ? Thanks in advance.

yunhailuo:
Thank you for your interests and PR, @Ayoob7 . This issue is a little bit special. The bottom line is it is not a bug but can use some improvements in coding and documentation.

If I remember it correctly, back in the day, GDC has some files named with only UUID which I found not useful to me and decided to rename them while downloading. So if you check the use of `get_ext`, you would find that they are used for constructing the new filename. Here is why I have this function and how I'd like it to work:

Say we have a file named "uuid.txt.tar.gz", I want to keep all the "txt.tar.gz" part and replace uuid with some meaningful name, say becoming "patient1_cancer.txt.tar.gz" not "patient1_cancer.gz". So get_ext should return "txt.tar.gz". This is important because if you unzip and extract the file, you can and should still have a "txt" extension. This is why I didn't simply keep the last postfix, "gz", but has `get_ext` to return that joint string. For what @arp95 has mentioned, say txt, tar and gz are not supported. So the filename doesn't have any supported format available, it will return the last postfix, gz, which is not "supported". Yes indeed but it is what I want. I want to get at least the last part rather than nothing because I want an extension no matter what. In another word, `get_ext` would check filename from left to right and return as soon as it found a recognized extension type and if none of them is recognized it should return the last one so that I have a relevant extension for file renaming rather than nothing. In this sense, I say it is not a bug because it functions as I want.

That being said, this issue can be improved. Either it should not be called "supported" but something else like "recognized" and documented clearly, or the whole file downloading and renaming part should be rewrite. In my opinion, those are directions this issue should go.

[...]

PR #8, opened 2019-03-09
Changed the get_ext file extension returning logic.

Ayoob7:
Changed the get_ext file extension returning logic. 

Thought of working on the file Downloading next, do you think it would be a good idea to use a nifty Python package called **PySpark**, if there is a possibility that we will need to implement Big Data (using Apache Spark) approaches to this server ( a server that can run across the shared resources of many computers) for large scale distributed processing?? 

Apache Spark is one of the best engines, for data heavy workloads and transformations to be done in distributed computing.

### DIFF
--- a/gdc.py
+++ b/gdc.py
@@ -18,7 +18,7 @@ import pandas as pd
 import requests
 
 GDC_API_BASE = 'https://api.gdc.cancer.gov'
-_SUPPORTED_FILE_TYPES = {'txt', 'vcf', 'bam', 'tsv', 'xml', 'maf', 'xlsx',
+_RECOGNIZED_FILE_TYPES = {'txt', 'vcf', 'bam', 'tsv', 'xml', 'maf', 'xlsx',
                          'tar', 'gz', 'md5', 'xls'}
 _SUPPORTED_DATASETS = [
         {'data_type': 'Copy Number Segment'},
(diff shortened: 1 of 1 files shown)


Ayoob7:
test cases : Passed
uuid.txt.tar1.gz => txt.gz
uuid.txt.tar.gz => txt.tar.gz
uuid.txt1.tar1.gz1 => gz1


Changed the get_ext file extension returning logic.

Thought of working on the file Downloading next, do you think it would be a good idea to use a nifty Python package called PySpark, if there is a possibility that we will need to implement Big Data (using Apache Spark) approaches to this server ( a server that can run across the shared resources of many computers) for large scale distributed processing??

Apache Spark is one of the best engines, for data heavy workloads and transformations to be done in distributed computing.

yunhailuo:
My original assumption is [GDC's filenames various a lot](https://portal.gdc.cancer.gov/repository?filters=%7B%22op%22%3A%22and%22%2C%22content%22%3A%5B%7B%22op%22%3A%22in%22%2C%22content%22%3A%7B%22field%22%3A%22files.data_format%22%2C%22value%22%3A%5B%22TXT%22%5D%7D%7D%5D%7D) but follow one rule which is having all metadata on the "left" and extensions all one the "right", or having metadata before extensions. For example, in "THETA_p_TCGA_Batch14_SNP_N_GenomeWideSNP_6_C03_455318.nocnv_grch38.seg.v2.txt", "THETA_p_TCGA_Batch14_SNP_N_GenomeWideSNP_6_C03_455318.nocnv_grch38.seg.v2" are all metadata and "txt" is an extension. Meanwhile, since this function is only used for renaming files, it doesn't matter that much whether this script recognize the extension or not, what matters more is I don't lose any extension while renaming even the script don't recognize some extension.

So for your first test case, "uuid.txt.tar1.gz", since txt is an extension rather than metadata, tar1.gz must also be extensions based on the assumption mentioned before. So it should be uuid.txt.tar1.gz => txt.tar1.gz; thus will be renamed as rename.txt.tar1.gz. This is what I mean by saying even the script doesn't recognize tar1 as an extension, it still preserves it based on the assumption and assuming tar1 is an unknown extension.

The reason I said it is not a bug in #1 is I didn't see the code breaks as long as my old assumption holds true. That being said, things could have changed and my old assumption may not be true. I suggest you study the filename on GDC a little bit, find some use cases and figure out how the original code breaks for those cases.

I'm not convinced about introducing PySpark. I may have the same suggestion, what some use cases would be here. Scripts for this repo is for data transformation and loading for Xena. In my opinion, the data transformation here is very basic and I haven't yet seen a need for advanced Big Data tool. And this repo is not for building any servers now or in my foreseeing future. I do respect your suggestions. So if you can list some potential use cases which may need or benefit from PySpark, that will be great. Thank you.

PR #12, opened 2019-03-11
Add travis CI and preliminary tests

ayan-b:
Related to #11

PR #13, opened 2019-03-12
Reorganized the project structure to be setuptools compatible.

yunhailuo:
Need more manual reviews and tests to make sure no obvious breaks. Then we can merge and test it more along the way.


requirements.txt
```diff
@@ -1,6 +0,0 @@
-requests==1.2.3
-numpy==1.13.0
-pandas==0.20.2
-jinja2==2.7.2
-lxml==4.0.0
-xlrd==1.1.0
```
ayan-b: We can keep requirements.txt (especially for those who do not want to install the package from pip)

setup.py
```diff
@@ -0,0 +1,41 @@
...
+
+SCRIPTS_DIR = os.path.join('xena_gdc_etl', 'scripts')
+scripts = glob.glob(os.path.join(SCRIPTS_DIR, '*.py'))
+scripts += glob.glob(os.path.join(SCRIPTS_DIR, '*.sh'))
+
+setup(
+    author='Yunhai Luo',
+    author_email='yunhail@stanford.edu',
+    classifiers=[
+        'Programming Language :: Python :: 2.7',
```
ayan-b: One part of [this](https://github.com/ucscXena/ucsc-xena-client/wiki/Google-Summer-of-Code-Ideas-2019#Update-GDC-data-ingestion-pipeline-and-run) project could be to make this code Python 3 compatible.

yunhailuo: Will add once we've really tested on python 3.0. I always kept to 2.7. Maybe the travis tests should be against python 3 in addition to python 2, @ayan-b ?

This is an effort we really should take. Moreover, I think we should, in the near future, stop worrying about python 2. Is this making sense to you, @maryjgoldman? I remember we talked about this before. It seems at least some other Xena python scripts are on 2.7 and we want this to be python 2 compatible?

ayan-b: @yunhailuo Yes, I have created travis.yml here https://github.com/yunhailuo/xena-GDC-ETL/pull/12 such that Python 2.7, 3.5, 3.6 & 3.7 are tested.

ayan-b: In my opinion, we should not worry anything older than 3.5 in 3.x versions.

requirements.txt
yunhailuo: My only worry is there are two copies and time goes by and these two are out of sync, which is not good.
An alternative could be to refer `requirement.txt` from within `setup.py`. But before that, I'm wondering what the use case could be for installing package without pip? Should the non-pip installer still goes into `setup.py` and resolve dependencies?

setup.py
yunhailuo: Sounds great. In your travis commit, for version you've tested, add all of them to setup.py. It makes sense to do it there even though this commit is supposed to setup `setup.py`.

requirements.txt
ayan-b: > My only worry is there are two copies and time goes by and these two are out of sync, which is not good.

We can use templates to create (i.e., setup.py.jj2 --> setup.py using Jinja2) setup.py and during CI we can test if they are in sync or not using git diff.

ayan-b: > But before that, I'm wondering what the use case could be for installing package without pip? Should the non-pip installer still goes into setup.py and resolve dependencies?

One may not want to install the whole package and use part of the code from this project in her project.

setup.py
ayan-b: Once this is merge, I will update the PR with the same.

requirements.txt
yunhailuo: > We can use templates to create (i.e., setup.py.jj2 --> setup.py using Jinja2) setup.py and during CI we can test if they are in sync or not using git diff.

Sounds too much/overkill to me. I'll just try to read `requirements.txt` in `setup.py`. I think it should work. But I'll have to wait until later tonight.

> One may not want to install the whole package and use part of the code from this project in her project.

Sure. That makes sense.

yunhailuo:
> PS: Will add more tests once this is merged along with working on the issues related to [this](https://github.com/ucscXena/ucsc-xena-client/wiki/Google-Summer-of-Code-Ideas-2019#Update-GDC-data-ingestion-pipeline-and-run) parallelly.

The test here is more or less just a placeholder. Please feel free to change it.

setup.py
```diff
@@ -0,0 +1,41 @@
...
+    tests_require=['pytest'],
+    long_description=long_description,
+    long_description_content_type='text/x-rst',
+    name='xena_gdc_etl',
+    packages=find_packages(),
+    package_data={'': ['resources/*']},
+    scripts=scripts,
+    url='https://github.com/yunhailuo/xena-GDC-ETL',
+    version='0.2.0',
+)
```
ayan-b: Maybe add keywords here?

yunhailuo:
* Spin out install_requires in requirements.txt. I would assume tests are mostly if not completely for us and there is no strong needs for just install partial and run tests. Let me know if a split is really required.
* Added keywords and license

ayan-b:
>     * Spin out install_requires in requirements.txt. I would assume tests are mostly if not completely for us and there is no strong needs for just install partial and run tests. Let me know if a split is really required.
> 
>     * Added keywords and license

Looks good. :+1. I think this PR is good to go now.

setup.py
ayan-b: @yunhailuo Should I point #12 's merging branch to this one?

yunhailuo: @ayan-b No need. Just merged this one. Thanks.

Issue #9, opened 2019-03-10
Release to pypi


ayan-b:
The project can be released to pypi so that it can be easily downloaded by pip or other installers.

yunhailuo:
That's a good suggestion. Since this is for Xena, I'd like to check with @maryjgoldman or @jingchunzhu to see if this is what we want to have.

ayan-b:
@yunhailuo also we need to add a LICENSE file. Which LICENSE does ucscxena follow?

yunhailuo:
This is a very legitimate suggestion. Again, @maryjgoldman or @jingchunzhu for proper license.

ayan-b:
@yunhailuo ,I think Apache 2.0 as per this site: http://xena.ucsc.edu/ .

maryjgoldman:
Yes, we use the Apache 2.0 license

pnijhara:
@ayan-b Are you still working on this issue? If not , I would like to work on this issue.

ayan-b:
Ideally maintainers should release to pypi.

pnijhara:
No issues.

yunhailuo:
> > @ayan-b Are you still working on this issue? If not, I would like to work on this issue.
> 
> Ideally maintainers should release to pypi.

I have no problem releasing it. As mentioned in my first comment, I'd like to have confirmation from @maryjgoldman or @jingchunzhu.

maryjgoldman:
Just realizing I never got back to you guys on this. I checked with Jing and she said this is fine to do.

yunhailuo:
> Just realizing I never got back to you guys on this. I checked with Jing and she said this is fine to do.

Thanks for checking. I'll do a test release and we can take a look first, @maryjgoldman and @ayan-b

yunhailuo:
Test release to be reviewed: https://test.pypi.org/project/xena-gdc-etl/

maryjgoldman:
I don't know enough about python to know if this works but the links above works and has a lot of great documentation. As we don't expect many people to interact with the code in this way, I vote we approve this and then if people complain then we can go back and change this.

yunhailuo:
One potential plan could be releasing this version at the end of GSoC as one end product, @maryjgoldman ?

maryjgoldman:
How does this work? Like if we change the code, will the pypi package update as well? Or is more like we release our code to pypi?

yunhailuo:
[...]

PR #14, opened 2019-03-12
Add LICENSE

ayan-b:
Related to https://github.com/yunhailuo/xena-GDC-ETL/issues/9


yunhailuo:
Added with GitHub template and fix year to 2017 so as to not expecting efforts on future changes. Thanks.

Issue #16, opened 2019-03-15
Improve CLI


ayan-b:
Currently, the user requires to type `python file_name args` in order to perform CLI operations. But, when the project will be installed using `setup.py`, the files will reside at `env/site_packages`, and thus will be hard for the end-user.

It can be improved by using a single CLI command, for example, `xena-gdc-etl gdc2xena args`
and `xena-gdc-etl gdc_check_new args`. In this way, the user needs not worry about path to the python files.

See also: https://github.com/yunhailuo/xena-GDC-ETL/pull/17#issuecomment-475110147

ayan-b:
@yunhailuo  Would like work on this once #13 is merged.

yunhailuo:
Sounds like a good feature to have. I just merged #13.

Two potential suggestions:

1. Feel free to prioritize scripts and we don't have to get all scripts in in one commit/PR. Actually better have a framework first then figure out specifics
2. Some clean up is going to be helpful. For example, I found `make_metadata.py` overlaps with `xena_dataset` module quite a lot. Such duplication is bad for maintenance. Also, some scripts may not be useful or relevant anymore.

ayan-b:
Ok, so I will do clean up first in another PR and then start working on improving CLI.

ayan-b:
Fixed in #23 & #24.

PR #17, opened 2019-03-18
Code cleanup

ayan-b:


ayan-b:
@yunhailuo  Submitted this PR to get quick feedback. :smiley:

yunhailuo:
Maybe rebase to master since I just merged #12 ?

ayan-b:
> Maybe rebase to master since I just merged #12 ?

Rebased. :)

ayan-b:
@yunhailuo, I think `equal_matrices`, `merge_xena`, `TARGET-CCSK_phenotype_ETL` modules could be inside `utils.py`.

ayan-b:
> @yunhailuo, I think `equal_matrices`, `merge_xena`, `TARGET-CCSK_phenotype_ETL` modules could be inside `utils.py`.

@yunhailuo What do you think about this? :)

yunhailuo:
`equal_matrices` and `merge_xena` can go to utils. I said they can go into utils only because they can be used by other python scripts/modules as well. But I would feel better if they are moved together with your #16 because they are, at least back in days for me, used for comparing Xena matrices, meaning an CLI interface seems necessary.

`TARGET-CCSK_phenotype_ETL`, if I remember correctly, is a really special case which is less likely to be applicable to other python scripts/modules. For me, it's better to leave it as a single, isolated, less used script.

yunhailuo:
BTW, I'm actually OK if you want to move one or a few functions at a time/PR, instead of doing everything perfect in one PR. The only recommendation is to make the PR message clear because I would, by default, do "Squash and merge" and use PR title as final commit title.

ayan-b:
In this particular PR, instead of `Squash and merge`, only `Merge` can be done since there are movements of different functions from different files(`mkdir_p`, `make_metadata`). I will separate out the commits in that way before merging.

yunhailuo:
> In this particular PR, instead of `Squash and merge`, only `Merge` can be done since there are movements of different functions from different files(`mkdir_p`, `make_metadata`). I will separate out the commits in that way before merging.
> 
> As a sidenote, pylint has a feature to check [code similarity](https://pylint.readthedocs.io/en/latest/technical_reference/features.html#similarities-checker). We can enable this in CI in the lint stage.

I would consider "Rebase and merge" first unless there is a strong reason for a merge commit. I don't want to deal with merge commits if I don't have to. Also, make sure you have a clean commit history if you don't want to have squash. I'll mark this PR as "Rebase and merge" for now.

yunhailuo:
Just a mark for myself, 15993f5 looks nice and clean by itself.

ayan-b:
@yunhailuo I think this PR is good to go. Let me know if some modification is required.

ayan-b:
@yunhailuo Should I make more changes? I don't think `make_metadata` overlaps anymore with `xena_datasets` module.

ayan-b:
> I'm not sure about your vision on make_metadata and xena_datasets. I just feel like it's better not to import scripts in a module.

Will follow.

ayan-b:
@yunhailuo I have found that quite a few modules inside scripts (namely `make_metadata`, `merge_xena`, `panTCGA`) `meta_templates` and `metavars` have been used. However, they are private data members inside `xena_dataset.GDCOmicset`. I feel it would be good if we declare them in `xena_gdc_etl/__init__.py` itself.

ayan-b:
Update on implementation: I have used stored the `METADATA_VARIABLES` and `METADATA_TEMPLATE` in a `constant.py` file instead of `__init__.py`.

yunhailuo:
> @yunhailuo I have found that quite a few modules inside scripts (namely `make_metadata`, `merge_xena`, `panTCGA`) `meta_templates` and `metavars` have been used. However, they are private data members inside `xena_dataset.GDCOmicset`. I feel it would be good if we declare them in `xena_gdc_etl/__init__.py` itself. In this way, it will available to every module which imports `xena_gdc_etl`. What do you think?

Sounds good. Please feel free to do that.

ayan-b:
It's already done. Instead of putting them inside `__init__.py`, I have put them in `constants.py`.

ayan-b:
@yunhailuo I think this PR can be merged. Let me know if some changes are required.

ayan-b:
> Hi. I see that the json module is imported but not used in tests/test_gdc.py.
It will be better if it's removed :)

Thanks for the review! Really appreciate it. :blush:

ayan-b:
I have marked some tests as `CI` as they require internet while executing.

yunhailuo:
> Note: I have marked some tests as `CI` as they require internet while executing. So, if someone is offline during coding, they can run `python -v -m "not CI"` in order to execute the tests which require no internet connection.

Just to check `python -v -m "CI"` is to execute tests requiring internet connection, right?

ayan-b:
Yup.

ayan-b:
@yunhailuo is the CONTRIBUTING guide OK or should remove/add something?

yunhailuo:
> @yunhailuo is the CONTRIBUTING guide OK or should I remove/add something?

Should be fine for now for this PR. Is this PR ready?

ayan-b:
This PR is ready and can be merged.

xena_gdc_etl/scripts/make_metadata.py
```diff
@@ -18,6 +18,11 @@
 
 import jinja2
 
+sys.path.insert(
+    0, os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+)
+from xena_gdc_etl.constants import METADATA_TEMPLATE, METADATA_VARIABLES # noqa
+
```
yunhailuo: Looks strange here (and below). Relative import doesn't work here?

ayan-b: Relative import works here also. I used this method in order to have a similar import structure in all files (i.e., inside `xena_GDC_ETL/Scripts/` and `xena_GDC_ETL/`). I will change this to relative import.

ayan-b:
> Since we are going to do rebase and merge for this PR, I'd appreciate it you could clean up the commit history a little bit. It seems ebb4970, 285da6b and 32ef0cc are dealing with one same issue. Should they be combined?

Sure.

ayan-b:
I have created #22 for the test fix since it does not really belong in this PR.

yunhailuo:
> I have created #22 for the test fix since it does not really belong in this PR.

That's good. Thanks again for the great work. I'll merge this one and see if #22 will pass.

PR #20, opened 2019-04-06
gdc.py: Show error code of failed download request

roshnet:


roshnet:
Yeah, I am working on it.
Thanks for the suggestion! 😃

roshnet:
As of now, the `/data` endpoint returns a 503 response, as the servers are down for maintenance.
Once the servers are up and running, we can expect a 404 response for the invalid resource the test case tries to access.

tests/test_gdc.py
```diff
@@ -69,6 +69,11 @@ def test_download():
     os.unlink(file_path)
 
 
+def test_download_error():
+    uuid = "something-invalid-0101"
+    gdc.download(uuid)
```
ayan-b: Great start! However, you are not asserting the output to some expected output, so the test is _not running_. You may take a look at the other tests in the same file for some inspiration. :smiley:

tests/test_gdc.py
```diff
@@ -88,3 +93,6 @@ def test_get_samples_clinical():
     project_id = "TCGA-OV"
     actual = gdc.get_samples_clinical(project_id)
     assert actual['case_id'][0] == "71faa2c1-0d5b-4dcc-bdf9-f2405f29907c"
+
+
+test_download_error()
```
ayan-b: You need not to execute the function. `pytest` will discover the test and will run by itself.

roshnet: Oops, I forgot to remove that part after testing locally, will remove it.

ayan-b:
@roshnet need a rebase.

PR #22, opened 2019-04-08
Add more test for gdc module

ayan-b:


yunhailuo:
Thanks for splitting this out of #17 .

PR #23, opened 2019-04-12
Code cleanup: Improve CLI

ayan-b:
- [X] `equal_matrices` (tests added)
- [X] `make_metadata` (tests added)
- [x] `merge_xena` (tests added)
- [x] `gdc_check_new`
- [ ] `gdc2xena`

Also set-up initial parser
Related to #16.

See also https://github.com/yunhailuo/xena-GDC-ETL/pull/17#issuecomment-475110147

PR #24, opened 2019-05-01
Add test for `gdc_check_new`

ayan-b:
- My previous implementation of `pytest.fixture` was wrong. In my implementation, I was wrongly creating fixtures not using them in the tests. 8c70670 removes the error.
- Also added the missing test for `gdc-check-new` which was not implemented in #23 in 5a5aa5a.

I will send a PR updating the docs once the CLI for `gdc2xena` is ready.

### DIFF
--- a/tests/test_gdc_check_new.py
+++ b/tests/test_gdc_check_new.py
@@ -5,18 +5,16 @@ except ImportError:
 
 import pandas as pd
 import pytest
-from mock import patch
 
 from xena_gdc_etl import gdc_check_new
 
 
 @pytest.mark.CI
-@patch("sys.stdout", new_callable=StringIO)
-def test_gdc_check_new(mocked_stdout):
+def test_gdc_check_new(capfd):
     url = "https://docs.gdc.cancer.gov/Data/Release_Notes/DR9.0_files_swap.txt.gz"  # noqa
     gdc_check_new.gdc_check_new(url)
-    actual = StringIO(mocked_stdout.getvalue())
-    actual = pd.read_csv(actual, sep='\t')
+    out, err = capfd.readouterr()
+    actual = pd.read_csv(StringIO(out), sep='\t')
     expected = pd.read_csv("tests/fixtures/gdc_check_new_DR9.0_files_swap.csv",
                            sep='\t')
     expected = expected.head()
--- a/tests/test_gdc_check_new.py
+++ b/tests/test_gdc_check_new.py
@@ -1,0 +1,23 @@
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
+
+import pandas as pd
+import pytest
+from mock import patch
+
+from xena_gdc_etl import gdc_check_new
+
+
+@pytest.mark.CI
+@patch("sys.stdout", new_callable=StringIO)
+def test_gdc_check_new(mocked_stdout):
+    url = "https://docs.gdc.cancer.gov/Data/Release_Notes/DR9.0_files_swap.txt.gz"  # noqa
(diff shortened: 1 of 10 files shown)


ayan-b:
@yunhailuo, for `gdc2xena` I think we can have a different entry point.

For example, currently we have the following CLI options:
```bash
xge make-metadata <args>
xge merge-xena <args>
...
```
however, for `gdc2xena`, we can have `gdc2xena <args>` instead of following it by `xge`.

What do you think?

tests/test_utils.py
```diff
@@ -8,10 +8,11 @@
     from io import StringIO
 
 import pandas as pd
-import pytest
 
 from xena_gdc_etl import utils
 
+PATH = "tests/fixtures/"  # path to static files
```
yunhailuo: Can we use https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions instead of this? I think that's a standard way to do shared fixtures, isn't it?

tests/test_gdc_check_new.py
```diff
@@ -0,0 +1,23 @@
...
+
+from xena_gdc_etl import gdc_check_new
+
+
+@pytest.mark.CI
+@patch("sys.stdout", new_callable=StringIO)
+def test_gdc_check_new(mocked_stdout):
+    url = "https://docs.gdc.cancer.gov/Data/Release_Notes/DR9.0_files_swap.txt.gz"  # noqa
+    gdc_check_new.gdc_check_new(url)
+    actual = StringIO(mocked_stdout.getvalue())
```
yunhailuo: Is this a more direct way to get stdout: https://docs.pytest.org/en/latest/capture.html ? Their example is not very clear. This might be a one: https://stackoverflow.com/questions/20507601/writing-a-pytest-function-for-checking-the-output-on-console-stdout-in-python

yunhailuo:
I think this is relatively focused. I'll squash and merge. Feel free to change the PR title if you'd like another commit message.

tests/test_utils.py
ayan-b: I tried using `conftest`, but it seems simply using a global variable `PATH` is somewhat more concise form.

yunhailuo: If `conftest` works. I'd prefer that. I worried having PATH may be incompatible with pytest in the long term. Unless `conftest` is much more complicated than PATH.

ayan-b: Actually, I don't understand what we should store in the `conftest` that will be shared between the functions.

yunhailuo: I'm thinking put `PATH` (or whatever constants) or load data files in conftest. I did get your point as those test data are used/input as file rather than loaded fixtures. My concern is your `PATH` is really like a shared constant while this `test_utils.py` is only one of many test files. Indeed, `PATH` is only used in this file. But if we use it in one or more other files in the future, this could be a problem for maintenance. Indeed, I'm not 100% sure how to do this. But `conftest` seems to be a pytest way for centralized data store and data sharing for multiple tests.

ayan-b: We could have a `PATH` global variable defined in `tests/utils.py`  and then we can set the `PATH` variable on top of each test file, e.g., `PATH = utils.PATH`. If we use a sub-directory, in that case, we can use something like, 
```python
PATH = utils.PATH + "xena_dataset"
```
What do you think about this approach?

yunhailuo: Well, still complicated. Let's keep it your old way. I realized it's in test. Maybe it's not ideal for maintenance but we won't miss it since tests will fail if something are out of sync.

PR #27, opened 2019-05-16
xena_dataset.py: Allow user to download certain number of files

ayan-b:


ayan-b:
@yunhailuo This PR is ready.

yunhailuo:
This doesn't seem to be very useful.

1. It's not very useful to just download 1 file as we discussed recently. You can test python code with 1 file but it's better to have a full set to really QA pipeline.
2. Though there is a bug related to private attribute name mangling and subclass, you could "relatively" easily (one or two lines with dict comprehension similar to your code?) modify `download_map` before calling `download`.

I'll try to fix the bug in #31

ayan-b:
> 1. It's not very useful to just download 1 file as we discussed recently. You can test python code with 1 file but it's better to have a full set to really QA pipeline.

Actually, you can download any number of files by passing `no_of_files=SOME_NUMBER`. This can be helpful when we may not want to download all the files associated with a datatype (due to file size etc.) but certain files in order to check if the resulted xena_matrices, metadata files are getting generated as expected.

> 2. Though there is a bug related to private attribute name mangling and subclass, you could "relatively" easily (one or two lines with dict comprehension similar to your code?) modify download_map before calling download.

Yes, that makes sense. :+1:

PR #32, opened 2019-05-27
Improve parser

ayan-b:
- Add test for `subcomm`.
- Remove `gdc2xena` as a different entry point instead incorporate it in the `xge` command: `xge etl`, `xge metadata`.
- Add `xge --version`.


yunhailuo:
I think this PR is good. But I'd like to hold on this and suggest a work on doc for command line. Can you take a look at README.rst and add a section about the command line. Potentially, it can be a section between "Basic usage with command line tools" and "Advanced usage with XenaDataset and its subclasses".

I'm trying to get an overall picture on how many commands/scripts we have and what can we do to organize them. For example,
* we have two entry points now, which might be what we have two. But I'm wondering if we could have just one?
* `equal_matrices` seems too simple to be a command. We might not want to expose it or not even have that function?
* `gdc_check_new` can be moved to utils.py?
* ...

ayan-b:
> - gdc_check_new can be moved to utils.py?
Moved

yunhailuo:
* I moved `gdc_check_new` to `gdc.py`. It's Xena independent thus should fit there
* `utils.metadata` is GDC-PANCAN specific. I don't think it's general enough to get exposed at CLI. Having it moved out of CLI, I don't think we will need subparsers to a subparser to a parser now. It could be something like:
  - `xge etl`
  - `xge metadata`
  - `xge xena-eql`
  - `xge gdc-check-new`
  - `xge merg-xena`

Please also check #36 and let me know if these look reasonable, @ayan-b .

ayan-b:
I think #36 needs to be merged first before I can start working on this.

yunhailuo:
Just merged #36.

ayan-b:
@yunhailuo, I have added another arg `--version`, i.e.,
```base
(env) ➜  xena-GDC-ETL git:(parser-test) xge --version
xge 0.2.0
```

ayan-b:
@yunhailuo, this PR is ready!

xena_gdc_etl/constants.py
```diff
@@ -1,3 +1,5 @@
+__version__ = "0.2.0"
```
yunhailuo: Why define it again here? Prone to out of sync.

Some strategies to maintain version number: https://packaging.python.org/guides/single-sourcing-package-version/
5 and 6 seems the simplest; feel free to choose your favorite though.
```
import pkg_resources
__version__ = pkg_resources.get_distribution('xena_gdc_etl').version
```

tests/test_main.py
```diff
@@ -9,21 +13,80 @@ def setUp(self):
...
+            "10",
+        ])
+        assert parsed.subcomm == "metadata"
+        assert parsed.project == "project_name"
+        assert parsed.datatype == "datatype"
+        assert parsed.matrix == "path/to/matrix"
+        assert parsed.release == 10.0
+
+
+def test_version_option():
```
yunhailuo: Doesn't really get it. Can you explain what is happening here?

tests/test_main.py
```diff
@@ -9,21 +13,80 @@ def setUp(self):
...
+    test_args = ["xge", "--version"]
+    with raises(SystemExit):
+        with patch.object(sys, "argv", test_args):
+            from xena_gdc_etl.main import main
+
+            main()
```
yunhailuo: Doesn't really get it. Can you explain what is happening here?

setup.py
```diff
@@ -5,6 +5,9 @@
 import os
 from setuptools import setup, find_packages
 
+
+__version__ = "0.2.0"
```
yunhailuo: Still need to define `__version__` here if you used `pkg_resources` below?

yunhailuo: I guess according to the Python doc, if you'd like to use `__version__`, you would use it in `__init__`?

ayan-b: Better if it is not defined. If it is defined in `__init__.py`, setup.py will try to import that will get `ImportError` for `from .xena_dataset import GDCOmicset`.

tests/test_main.py
ayan-b: We cannot test this as done previously (i.e., using `parse` since this works by raising an exception (SystemExit) and then exiting the program). So here we are checking if SystemExit is thrown and that particular code block is substituted when `xge --version` is run.

yunhailuo: Thanks. That helps. I feel it's a little confusing/complicated. The following seems to work. But I'm not sure it fits what you'd like to have here?
```
    @pytest.fixture(autouse=True)
    def _capture(self, capfd):
        self.capfd = capfd

    def test_version(self):
        with raises(SystemExit):
            self.parser.parse_args(['--version'])
        out, _ = self.capfd.readouterr()
        __version__ = pkg_resources.get_distribution("xena_gdc_etl").version
        self.assertEqual(out, 'xge {}\n'.format(__version__))
```

ayan-b: @yunhailuo Sorry for the delay. Modified the `--version` test.

yunhailuo:
Looks good for rebase and merge. Let me know when you think it's ready, @ayan-b .

ayan-b:
This PR is ready!

PR #33, opened 2019-05-28
CI: Test for min requirements

ayan-b:
This PR checks for the minimum requirements in as mentioned in the README file.

Since we are specifying some minimum versions for the dependencies in the README file, I thought it would be a good idea to check for those minimum versions in the CI. If in future, we unknowingly use functions from the dependencies that are not available in the older versions, that would be problematic.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,6 @@ script:
   - pip install .
   - pip freeze
 
-before_install:
+before_script:
   - pip install -r requirements.txt
   - pip install pytest~=3.6.1 pytest-cov~=2.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,5 +56,4 @@ script:
   - pip freeze
 
 before_script:
-  - pip install -r requirements.txt
   - pip install pytest~=3.6.1 pytest-cov~=2.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ script:
 before_install:
   - if [[ $TRAVIS_PYTHON_VERSION == "2.6" ]]; then pip install flake8==2.6.2; fi
   - if [[ -f min-requirements.txt && "$MINREQ" -eq 1 ]]; then
-      mv min_requirements.txt requirements.txt;
+      mv min-requirements.txt requirements.txt;
     fi
   - pip install -r requirements.txt
   - pip install pytest~=3.6.1 pytest-cov~=2.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,6 @@ script:
   - pip freeze
 
 before_install:
-  - if [[ $TRAVIS_PYTHON_VERSION == "2.6" ]]; then pip install flake8==2.6.2; fi
   - if [[ -f min-requirements.txt && "$MINREQ" -eq 1 ]]; then
       mv min-requirements.txt requirements.txt;
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,22 @@ jobs:
     - *lint
     - python: 3.6
       env: MINREQ=1
+      <<: *disable_global
+      script:
+        - pip install -r min-requirements.txt
+        - pip install pytest~=3.6.1 pytest-cov~=2.4
+        - pytest --cov=xena_gdc_etl
+        - pip install .
+        - pip freeze
     - python: 2.7
       env: MINREQ=1
+      <<: *disable_global
+      script:
+        - pip install -r min-requirements.txt
+        - pip install pytest~=3.6.1 pytest-cov~=2.4
+        - pytest --cov=xena_gdc_etl
+        - pip install .
+        - pip freeze
   allow_failures:
     - *lint
 
@@ -42,8 +56,5 @@ script:
   - pip freeze
 
 before_install:
-  - if [[ -f min-requirements.txt && "$MINREQ" -eq 1 ]]; then
-      mv min-requirements.txt requirements.txt;
-    fi
   - pip install -r requirements.txt
   - pip install pytest~=3.6.1 pytest-cov~=2.4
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -29,6 +29,8 @@ stages:
 jobs:
   include:
     - *lint
+    - python: 2.7
+      env: MINREQ=1
```
yunhailuo: What is this for?

ayan-b: `env: MINREQ=1` sets the `MINREQ` environment variable in that particular job.

.travis.yml
```diff
@@ -37,8 +39,12 @@ stage: test
 script:
   - pytest --cov=xena_gdc_etl
   - pip install .
+  - pip freeze
 
 before_install:
   - if [[ $TRAVIS_PYTHON_VERSION == "2.6" ]]; then pip install flake8==2.6.2; fi
+  - if [[ -f min-requirements.txt && "$MINREQ" -eq 1 ]]; then
```
yunhailuo: I guess I'm trying to ask why do you need that?

yunhailuo:
I think what I was trying to say is maybe we should only use minimum requirements in test?

ayan-b:
> I think what I was trying to say is maybe we should only use minimum requirements in test?

One possible downside is: if someone has a newer version in their system, in that case, they still need to uninstall the newer one and install the older one.

yunhailuo:
If we use `requirements.txt`, it will have version fixed. But if we use just `pip install`, it will use newer versions, won't it?

ayan-b:
> If we use `requirements.txt`, it will have version fixed. But if we use just `pip install`, it will use newer versions, won't it?

If we use `==` it will install that exact version, `>=` will install the latest version.

ayan-b:
What I think, we should always use the latest version so that we have the latest and greatest features (that are supported by the minimum version as well), but if we want to use a new feature that is only available in the latest version, we will simply bump up the minimum version.

yunhailuo:
> > If we use `requirements.txt`, it will have version fixed. But if we use just `pip install`, it will use newer versions, won't it?
> 
> If we use `==` it will install that exact version, `>=` will install the latest version.

OK. Good. Here is what I think:
We can test every possible combinations. So I'd say we clean out test to:
* Highest Python 2 (2.7) with fixed version
* Highest Python 2 (2.7) with latest version
* Highest Python 3 (non-dev) with fixed version
* Highest Python 3 (non-dev) with latest version

And don't worry about other versions for now. We can add later if someone ask for.

yunhailuo:
> What I think, we should always use the latest version so that we have the latest and greatest features (that are supported by the minimum version as well), but if we want to use a new feature that is only available in the latest version, we will simply bump up the minimum version.

We can use just `==` in requirements. That also works.

ayan-b:
@yunhailuo Done! :smiley:

.travis.yml
yunhailuo: Again, I'm not sure about the logic here. Can you elaborate a little bit? Can you just go head and test instead of using if-then?

ayan-b: The logic is: If there is `min-requirements.txt` file and `MINREQ` is set then move `min-requirements.txt` to `requirements.txt`.

yunhailuo: Why not just install `min-requirements`?

ayan-b: Like this https://github.com/ayan-b/xena-GDC-ETL/blob/travis/.travis.yml?

.travis.yml
```diff
@@ -29,6 +27,24 @@ stages:
 jobs:
   include:
     - *lint
+    - python: 3.6
+      env: MINREQ=1
```
yunhailuo: Do you still need `MINREQ` here and below?

ayan-b: Helps to identify the corresponding job.

yunhailuo:
1. Do you know why requirements are installed twice: https://travis-ci.com/yunhailuo/xena-GDC-ETL/jobs/204304313#L443-L513?

2. Should `pip install .` be independent so that it could check installation without pre-install requirements?

3. Is there something we can to to better organize 4 jobs (except lint) in tests? Two are now run with global setting and two are run without. It's not straight forward to understand. But I'm not expert on tests either. One way I'm looking at possible combination:
* `pip install .` with Python 3 without pre-installing requirements since it uses requirements anyway;
* `pip install .` with Python 2 without pre-installing requirements since it uses requirements anyway;
* `pip install -r min-requirements.txt` with following installments and tests in Python 3
* `pip install -r min-requirements.txt` with following installments and tests in Python 2
* Lint tests

4. It feels really wrong to allow failure on tests. Maybe we should fix lint, though it's not a priority at all.

.travis.yml
yunhailuo: Is this the way to go?
https://docs.travis-ci.com/user/build-stages/#naming-your-jobs-within-build-stages

yunhailuo: Otherwise we can make it clearer like `MIN_REQ_TEST=True` or something.

min-requirements.txt
```diff
@@ -0,0 +1,7 @@
+jinja2 == 2.7.2
+lxml == 4.0.0
+numpy == 1.13.0
+pandas == 0.20.2
+requests == 2.20.0
+xlrd == 1.1.0
+mock
```
yunhailuo: A minimum version for this one?

ayan-b: Done!

yunhailuo:
I think makefile might be overkill here, though I'm not sure. I'll merge this PR and play on top of it a little bit myself. Nice job!

ayan-b:
Renaming left!

yunhailuo:
> Renaming left!

That's fine. The goal of this PR is achieved. Simplify/reorganizing tests can be another PR.

ayan-b:
This PR is ready. :tada:

.travis.yml
ayan-b: Addressed in dc21bad.

PR #34, opened 2019-05-28
Add API method to generate Phenotype Matrix

ayan-b:


ayan-b:
@yunhailuo Please review.

ayan-b:
Possibly out of scope here, I also feel `gdc.py` can be a separate module altogether (maybe named as `pyGDC` or something like that`) so that others can use it.

ayan-b:
Failing test fixed in #40.

yunhailuo:
> Failing test fixed in #40.

I'm not sure why the test would be fixed in another PR. It's caused by this PR, right?

ayan-b:
> > Failing test fixed in #40.
> 
> I'm not sure why the test would be fixed in another PR. It's caused by this PR, right?

No, data was updated in GDC, so it was failing.

yunhailuo:
I've been thinking about and started to feel maybe request on File endpoint and get info through `cases` might work as long as it's not too big for memory:

1) Filter:
```
{
    "files.access": "open",
    "files.data_category": "Sequencing Reads",
}
```
2) Expand (based on Mary's list; you can do fields if you want to):
```
[
    'cases.demographic',
    'cases.diagnoses',
    'cases.exposures',
    'cases.family_histories',
    'cases.project',
    'cases.samples',
    'cases.samples.portions',
    'cases.samples.portions.slides',
    'cases.tissue_source_site',
]
```
3) One you got the JSON, you got a list of dict. You can transform it into a big dict which will be keyed on sample ID. Every item of this dict is a dict whose keys are in Mary's list. While transforming, you can skip samples if their ID already exists.

ayan-b:
So, what I am doing:
1. Get all the sample ids from `files` endpoint.
2. Get all the fields/expands from `cases` endpoint using those sample ids.

yunhailuo:
> So, what I am doing:
> 
> 1. Get all the sample ids from `files` endpoint.
> 2. Get all the fields/expands from `cases` endpoint using those sample ids.

That also works. But wouldn't it be two (actually four) calls rather than one (actually two) call?

yunhailuo:
This is [an example](https://api.gdc.cancer.gov/files/?size=1&expand=cases.demographic,cases.diagnoses,cases.exposures,cases.family_histories,cases.project,cases.samples,cases.samples.portions,cases.samples.portions.slides,cases.tissue_source_site&filters=%7B%22op%22%3A%22and%22%2C%22content%22%3A%5B%7B%22op%22%3A%22in%22%2C%22content%22%3A%7B%22field%22%3A%22cases.project.project_id%22%2C%22value%22%3A%5B%22TCGA-CHOL%22%5D%7D%7D%2C%7B%22op%22%3A%22in%22%2C%22content%22%3A%7B%22field%22%3A%22files.data_category%22%2C%22value%22%3A%5B%22Sequencing%20Reads%22%5D%7D%7D%5D%7D) for one (actually the first of the two; size=1) call request.

ayan-b:
By this comment https://github.com/yunhailuo/xena-GDC-ETL/pull/34#issuecomment-499757536, you mean filter something like this?
```json
{
    "op":"and",
    "content":[
        {"op":"in", "content":{"field": "files.data_category", "value":"Sequencing Reads"}},
        {"op":"in","content":{"field":"cases.project.project_id","value":["TCGA-CHOL"]}},
        {"op":"in","content":{"field":"files.data_category","value":["Sequencing Reads"]}}
        ]
}
```

yunhailuo:
> By this comment [#34 (comment)](https://github.com/yunhailuo/xena-GDC-ETL/pull/34#issuecomment-499757536), you mean filter something like this?
> 
> ```json
> {
>     "op":"and",
>     "content":[
>         {"op":"in","content":{"field":"cases.project.project_id","value":["TCGA-CHOL"]}},
>         {"op":"in","content":{"field":"files.data_category","value":["Sequencing Reads"]}}
>         ]
> }
> ```

Right. It should work. I'm wondering if you need `files.data_category`. Is `data_category` itself a field for the File endpoint?

ayan-b:
> I'm wondering if you need files.data_category. Is data_category itself a field for the File endpoint?

Not really sure. But, a couple of `submitter.sample_ids` field gets vacant if data category is not included in the filter.

yunhailuo:
> > I'm wondering if you need files.data_category. Is data_category itself a field for the File endpoint?
> 
> Not really sure. But, a couple of `submitter.sample_ids` field gets vacant if data category is not included in the filter.

OK. I mean if this will work:
```
{
    "op":"and",
    "content":[
        {"op":"in","content":{"field":"cases.project.project_id","value":["TCGA-CHOL"]}},
        {"op":"in","content":{"field":"data_category","value":["Sequencing Reads"]}}
        ]
}
```

ayan-b:
@yunhailuo, Yes, it works.

yunhailuo:
Great. Is it now clear what you could do? Or you need more clarification?

ayan-b:
Will add more detailed tests once the implementation is approved.

ayan-b:
@yunhailuo Please review.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,169 +1471,221 @@ def transform(self):
...
                 )
-            elif all([i.startswith('TARGET-') for i in self.projects]):
-                xena_matrix = api_clin.dropna(axis=1, how='all').set_index(
-                    'submitter_id.samples'
+                xena_matrix.set_index(
+                    "cases.samples.submitter_id",
+                    inplace=True,
                 )
+            elif all([i.startswith("TARGET-") for i in self.projects]):
+                raise NotImplementedError
```
yunhailuo: This part above is great!
For TARGET projects, just don't allow this `xena_dtype`. As mentioned above, put it as a new `xena_dtype` instead of using `legacy`. It's not really a legacy but just a different ETL strategy (meaning API only vs API+XML).

ayan-b: For TARGET projects, generating matrix from API will be implemented in this week I guess.

ayan-b:
@yunhailuo, please review.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1344,47 +1348,56 @@ def gdc_filter(self, filter_dict):
...
-                    + '.'
-                    + file_df['file_id'].astype(str)
-                    + '.'
-                    + file_df['file_name'].apply(gdc.get_ext)
-                ).to_dict()
-            if not file_dict:
-                msg = '\rNo {} data found for project {}.'
-                gdc_dtype = self._XENA_GDC_DTYPE[self.xena_dtype]
+                assert self._download_map
+                return self._download_map
```
yunhailuo: This will prevent `download_map` being overwritten by users. The logic behind this getter is to provide a default value. By saying default, it means it can be changed.

I have to say I'm not 100% sure using private property is the correct way to do this. I got it online and it seems fine. Let me know if you have a better algorithm though.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1344,47 +1348,56 @@ def gdc_filter(self, filter_dict):
 
     @XenaDataset.download_map.getter
     def download_map(self):
-        try:
-            assert self._download_map
-            return self._download_map
-        except (AttributeError, AssertionError):
-            fields = ['file_id', 'file_name', 'data_category']
+        if self.xena_dtype == "Xena_phenotype":
+            return {}
```
yunhailuo: Please add a message here (and after this if proper) showing nothing will be (or was) downloaded.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1344,47 +1348,56 @@ def gdc_filter(self, filter_dict):
...
+                }
+                self._download_map = file_dict
+                msg = '\r{} files found for {} data of {}.'
                 print(
                     msg.format(
-                        ' - '.join(sorted(gdc_dtype.values())),
-                        str(self.projects),
+                        len(file_dict),
+                        self.xena_dtype,
+                        self.projects,
```
yunhailuo: OK in one line? At least not style change in this PR please. Confusing for code review.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
-                    cases_samples_map,
-                    how='inner',
-                    on='TARGET USI',
-                ).set_index('sample_id')
+                xena_matrix.set_index(
+                    "cases.samples.submitter_id",
+                    inplace=True,
+                )
+            elif all([i.startswith("TARGET-") for i in self.projects]):
+                raise NotImplementedError
```
yunhailuo: Here again. If you really want to do `NotImplementedError`, please add a proper message with you `raise`.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
             else:
-                warnings.warn('Fail to get per sample based clinical matrix.')
-        elif self.xena_dtype == 'biospecimen':
+                raise ValueError(
+                    'Getting "GDC_phenotype" for a cohort with mixed TCGA '
+                    'and TARGET projects is not currently suppported.'
+                )
```
yunhailuo: `GDC_phenotype` or `Xena_phenotype`.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
+        else:
+            total = len(self.raw_data_list)
```
yunhailuo: Since you will `return` or `raise` in the if above, there is no need for an else and indent for the code block below. For one thing, I personally prefer less indentation/nesting. Besides, you will have a cleaner diff here so that it may help us see changes more clear (hopefully less confusions like `GDC_phenotype` vs `Xena_phenotype` above).

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
+            count = 0
+            bio_dfs = []
+            clin_dfs = []
+            for path in self.raw_data_list:
+                count = count + 1
+                print('\rProcessing {}/{} file...'.format(
+                    count, total
+                ), end='')
```
yunhailuo: Same here, please no changes here if not absolutely needed (i.e. no style changes or indentation if possible). The diff can be much cleaner and easy for me. Thank you!

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
+                sys.stdout.flush()
+                # `read_biospecimen` and `read_clinical` will check file
+                # format
```
yunhailuo: Same here, please no changes here if not absolutely needed (i.e. no style changes or indentation if possible). The diff can be much cleaner and easy for me. Thank you!

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
+                    df = read_clinical(path)
+                    clin_dfs.append(df)
+                except Exception:
+                    try:
+                        df = read_biospecimen(path)
+                        bio_dfs.append(df)
+                    except Exception:
+                        raise TypeError('Fail to process file {}.'.format(
+                            path
+                        ))
```
yunhailuo: Same here, please no changes here if not absolutely needed (i.e. no style changes or indentation if possible). The diff can be much cleaner and easy for me. Thank you!

I think more below but I won't label them. Please help me check.

xena_gdc_etl/xena_dataset.py
ayan-b: Goes 79+ characters.

xena_gdc_etl/xena_dataset.py
ayan-b: 79+ characters here as well.

xena_gdc_etl/xena_dataset.py
ayan-b: Same 79+ and flake8 will start complaining.

xena_gdc_etl/xena_dataset.py
ayan-b: Goes 79+ characters.

xena_gdc_etl/xena_dataset.py
yunhailuo: Well, not if "no style changes or *indentation* if possible", right? It passes lint before...

Mentioned in previous comment, I don't think else and indentation is necessary...

xena_gdc_etl/xena_dataset.py
ayan-b: If I `return self` that will lead to code duplication of ~4 lines.

yunhailuo: Sorry. Let me make it more clear about what "surgery" I had in my mind. Commented below.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
-            # `read_biospecimen` and `read_clinical` will check file format
-            try:
-                df = read_clinical(path)
-                clin_dfs.append(df)
-            except Exception:
-                try:
-                    df = read_biospecimen(path)
-                    bio_dfs.append(df)
-                except Exception:
-                    raise TypeError('Fail to process file {}.'.format(path))
```
yunhailuo: `self.raw_data_list` is empty. And it's reasonable to run the `for` here since `download` is run and the pattern would be kept here.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
-        print('\rAll {} files have been processed. '.format(total))
```
yunhailuo: `self.raw_data_list` is empty and it's reasonable to run `for` here since `download` is run and the pattern would be kept here. No file will be processed. The message is fine unless you think it has to be specific and clear.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
-            bio_matrix = pd.DataFrame()
-        try:
-            clin_matrix = (
-                pd.concat(clin_dfs, axis=0)
-                .replace(r'\r\n', ' ', regex=True)
-                .replace(r'^\s*$', np.nan, regex=True)
-                .dropna(axis=1, how='all')
-            )
-        except Exception:
-            clin_matrix = pd.DataFrame()
```
yunhailuo: Both `try` would fail and gives empty `bio_matrix` and `clin_matrix` if nothing is done. I think it's fine to let them fail but I would also agree if you think it's not a good or readable logic for the future. Overall, this is a patten following the `download` and the `for` loop above. That's why I'm OK running it. Feel free to change it if you want.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1459,170 +1472,222 @@ def transform(self):
...
+                    pd.concat(clin_dfs, axis=0)
+                    .replace(r'\r\n', ' ', regex=True)
+                    .replace(r'^\s*$', np.nan, regex=True)
+                    .dropna(axis=1, how='all')
                 )
-                .replace(r'^\s*$', np.nan, regex=True)
-                .set_index('bcr_sample_barcode')
-                .fillna(bio_matrix.set_index('bcr_sample_barcode'))
-            )
-        if self.xena_dtype == 'GDC_phenotype':
```
yunhailuo: This is how I deal with API before. Probably here is a reason why you think it's a legacy. I still think it's a new type/strategy. So I would imagine an `elif` here for `Xena_phenotype` and output a `xena_matrix` for the last five lines.

yunhailuo:
Overall, my personal guideline is to

1. Very first and important, make all reasonable changes if needed;
2. Make less changes if possible so that it's easy for coding and review, and hopefully with less confusions or mistakes.

ayan-b:
Thanks! This clearly looks much cleaner. Please review now @yunhailuo.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1215,8 +1216,11 @@ class GDCPhenoset(XenaDataset):
             'data_category': ['Biospecimen', 'Clinical'],
             'data_format': 'BCR XML',
         },
+        'Xena_phenotype': {
+            'data_category': ['Biospecimen', 'Clinical'],
```
yunhailuo: Should this be empty?

xena_gdc_etl/xena_dataset.py
```diff
@@ -1344,6 +1348,12 @@ def gdc_filter(self, filter_dict):
 
     @XenaDataset.download_map.getter
     def download_map(self):
+        if self.xena_dtype == "Xena_phenotype":
+            print(
+                "'Xena_phenotype' datatype is selected, no files will be "
+                "downloaded"
+            )
+            return {}
```
yunhailuo: Still force default and not allow objects to overwrite download map?

tests/test_xena_dataset.py
```diff
@@ -18,3 +21,18 @@ def test_read_clinical():
...
+            dataset = GDCPhenoset(
+                projects=["TCGA-CHOL", "TARGET-WT"],
+                xena_dtype="Xena_phenotype",
+            )
+            dataset.download().transform()
+        error_str = (
+            "Getting 'Xena_phenotype' for a cohort with mixed TCGA "
+            "and TARGET projects is not currently suppported."
+        )
+        assert exception_info.value.args[0] == error_str
```
yunhailuo: If you are going to:

> Will add more detailed tests once the implementation is approved.

don't add just one here, especially not a test on error message...

ayan-b: I thought this error message is the safest one to add tests (i.e, less prone to change). :/

xena_gdc_etl/xena_dataset.py
ayan-b: Isn't data category biospecimen and clinical (or sequencing reads) ?

tests/test_xena_dataset.py
yunhailuo: But will you add more tests or not? If yes, add all of them in the future PR. No need to single one out.

xena_gdc_etl/xena_dataset.py
yunhailuo: If I remember this correctly (and you should also check), this is for downloading files. Since nothing will be downloaded, I think this should be an empty dict.

xena_gdc_etl/xena_dataset.py
ayan-b: Done.

ayan-b:
Thanks! Its already in the slack.

ayan-b:
Planning to use the code of this PR for `CPTAC-3` phenotype data.

ayan-b:
Closing in favor of #91.

PR #36, opened 2019-05-30
Expose useful classes and organize code

yunhailuo:
* Expose `XenaDataset`, `GDCOmicset`, `GDCPhenoset` and `GDCSurvivalset`;
* `gdc_check_new` can be a GDC specific function, independent of Xena;
* The old `utils.metadata` is actually specific for "GDC-PANCAN"; move it to `panTCGA.py` for now.


xena_gdc_etl/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .xena_dataset import (
+    XenaDataset,
+    GDCOmicset,
+    GDCPhenoset,
+    GDCSurvivalset,
+)
```
ayan-b: Maybe append `# noqa` here so that `flake8` doesn't complain? :sweat_smile:

yunhailuo: Thanks.

PR #37, opened 2019-05-31
Lint fixes

ayan-b:


ayan-b:
@yunhailuo This PR is ready. Please review. :blush:

yunhailuo:
Sorry. Just re-organized travis code. Please rebase.

setup.cfg
```diff
@@ -5,4 +5,5 @@
 # W605: invalid escape sequence '\*'
 # E121: continuation line under-indented for hanging indent
 # E126: continuation line over-indented for hanging indent
-ignore = E722, W605, E121, E126
+# W503: line break before binary operator
+ignore = E722, W605, E121, E126, W503
```
yunhailuo: Does this mean ever `#noqa` will ignore these five no matter what?

ayan-b: Yup, flake8 will ignore those errors regardless of `#noqa`.

ayan-b:
Rebased. :)

ayan-b:
Please take a look @yunhailuo.

yunhailuo:
I have some hesitation on how we would ignore rules. Like [W503](https://lintlyci.github.io/Flake8Rules/rules/W503.html) should be ignored as they said so. But others, overall ignores in config or inline ignore, I'd prefer to be careful and specific. By being specific, I mean
* Overall ignore should be restricted unless there is clear reason to do so;
* Inline ignore should be specific to certain rule(s).

This is my general thinking. We will need some discussions like:
* Which rule(s) should definitely be ignored for all? Only W503?
* Is it overwhelming if for some rules, which are not really ignore for all, are too many for inline?
* ...

xena_gdc_etl/constants.py
```diff
@@ -1,3 +1,5 @@
+GDC_RELEASE_URL = 'https://docs.gdc.cancer.gov/Data/Release_Notes/Data_Release_Notes/#'  # noqa: E501
```
yunhailuo: Doesn't seem it should have hash tag at the end?

ayan-b: The hashtag has been used so that we can simply append `data-release-N` and it will point to that Nth release.

yunhailuo: The hash tag indicates anchor and should be together with the tag rather than the URL.

yunhailuo: I think to be exact, # is neither part of the URL nor part of the identifier according to [the URL writing standard](https://url.spec.whatwg.org/#url-writing). But the standard for URL class in API put # part of the hash attribute with the identifier rather than with the part in front of it: https://url.spec.whatwg.org/#dom-url-hash

ayan-b: Done!

yunhailuo:
Looks good. Let me know your thoughts on global ignore when you are ready.

ayan-b:
[WIP Comment]
Global Ignores:
- **W605 (invalid escape sequence '\*')**: Quite a few docstrings use `*/*/`, so its ignored.
- **E722 (do not use bare 'except')**:
- **E121: continuation line under-indented for hanging indent**
- **E126: continuation line over-indented for hanging indent**
- **W503: line break before binary operator**

yunhailuo:
> * Quite a few docstrings use `*/*/`

An example?

ayan-b:
> > * Quite a few docstrings use `*/*/`
> 
> An example?

https://github.com/yunhailuo/xena-GDC-ETL/blob/76b792ca984be32874687a3d682b3e06bdd95a36/xena_gdc_etl/xena_dataset.py#L430

ayan-b:
51ecf1a uses `black` to format code: `black -l 79 -S xena_gdc_etl`.

yunhailuo:
Thanks for the example. Looked into W605 a little bit. In the long run, it may become SyntaxError. We should not ignore it. [It seems make docstring a raw string will fix it.](https://github.com/PyCQA/pycodestyle/issues/854) I haven't try though. I saw only four in `xena_dataset.py`.

yunhailuo:
> [02b9ead](https://github.com/yunhailuo/xena-GDC-ETL/commit/02b9ead875e133f825c82c174b2b460828d1663f) uses `black` to format code: `black -l 79 -S xena_gdc_etl tests`.

What is black? A link for intro or doc?

xena_gdc_etl/constants.py
```diff
@@ -1,4 +1,6 @@
-GDC_RELEASE_URL = 'https://docs.gdc.cancer.gov/Data/Release_Notes/Data_Release_Notes/'  # noqa: E501
+GDC_RELEASE_URL = (
+    'https://docs.gdc.cancer.gov/Data/Release_Notes/Data_Release_Notes/'
+)  # noqa: E501
```
yunhailuo: Still need to ignore 501?

ayan-b:
@yunhailuo black: https://black.readthedocs.io/

xena_gdc_etl/gdc.py
```diff
@@ -291,10 +325,11 @@ def download(uuids, download_dir='.', chunk_size=4096):
             file_size = int(response.headers['Content-Length'])
             if uuids[uuid] is None:
                 content_disp = response.headers['Content-Disposition']
-                ori_name = content_disp[content_disp.find('filename=') + 9:]
+                ori_name = content_disp[content_disp.find('filename=') + 9 :]  # noqa: E203, E501
```
yunhailuo: Why putting a space and triggering E203 here?
Still need E501?

xena_gdc_etl/gdc.py
```diff
@@ -382,12 +438,14 @@ def get_samples_clinical(projects=None):
     # correctly with ``record_path `` "samples". Use the raw json instead.
     # Besides, there are cases (34 by 12/11/2017) which doesn't have any
     # samples and thus doesn't have the key "samples". Ignore them.
-#    for r in res:
-#        r.setdefault('samples', [{}])
-#        samples_json.append(r)
+    #    for r in res:
+    #        r.setdefault('samples', [{}])
+    #        samples_json.append(r)
```
yunhailuo: Should we just delete these three lines? I'd like to but need a second opinion.

xena_gdc_etl/gdc.py
```diff
@@ -399,14 +457,19 @@ def gdc_check_new(new_file_uuids):
     """
 
     df_list = []
-    for uuids in (new_file_uuids[i:i + 20000]
-                  for i in range(0, len(new_file_uuids), 20000)):
+    for uuids in (
+        new_file_uuids[i : i + 20000]  # noqa: E203
```
yunhailuo: Again, but I'm wondering now why it keeps on doing this? I assume `black` does this?

xena_gdc_etl/main.py
```diff
@@ -97,123 +109,181 @@ def create_parser():
...
+        "df2", type=str, help='Directory for the second matrix.'
     )
     # gdc_check_new subparser
     gdc_check_new_parser = subparsers.add_parser(
         "gdc-check-new",
         description="Check GDC's list of updated files and summarize "
-                    "impacted project(s), data_type(s) and "
-                    "analysis.workflow_type(s)."
+        "impacted project(s), data_type(s) and "
+        "analysis.workflow_type(s).",
```
yunhailuo: Wondering if this is a right/preferred style here and below, since `black` does something "strange"...

ayan-b:
No longer ignoring `W605`.

xena_gdc_etl/xena_dataset.py
```diff
@@ -59,9 +59,11 @@ def read_by_ext(filename, mode='r'):
     ext = os.path.splitext(filename)[1]
     if ext == '.gz':
         import gzip
+
```
yunhailuo: It's probably not preferred this way. But it happens when I don't want to just import no matter what.
My real point here is no need for a new line at least to flake8, right?

xena_gdc_etl/xena_dataset.py
```diff
@@ -59,9 +59,11 @@ def read_by_ext(filename, mode='r'):
...
         return gzip.open(filename, mode)
     elif ext == '.bz2':
         import bz2
+
```
yunhailuo: Doesn't need a new line for flake8, right?

xena_gdc_etl/xena_dataset.py
```diff
@@ -130,9 +136,7 @@ def read_biospecimen(fileobj):
         record.update(samples_common)
         samples[record['bcr_sample_barcode']] = record
     df = pd.DataFrame(samples).T
-    sample_mask = df.bcr_sample_barcode.map(
-        lambda s: s[-3:-1] not in ['10']
-    )
+    sample_mask = df.bcr_sample_barcode.map(lambda s: s[-3:-1] not in ['10'])
```
yunhailuo: Interesting. Not having a space before colon here...

xena_gdc_etl/xena_dataset.py
```diff
@@ -663,9 +710,11 @@ def metadata(self):
...
             assert os.path.isfile(self.matrix)
         except AttributeError:
-            raise IOError('Xena matrix for this dataset is unknown; please '
-                          'create a matrix or assign an existing matrix file '
-                          'to the "matrix" property before making metadata.')
+            raise IOError(
+                'Xena matrix for this dataset is unknown; please '
+                'create a matrix or assign an existing matrix file '
+                'to the "matrix" property before making metadata.'
+            )
```
yunhailuo: For string like this (help message for CLI above, etc), we can reset the line break to use more of the 79 limit.
Not very critical, though.

xena_gdc_etl/gdc.py
ayan-b: PEP8 recommends to put spaces around `:` during slicing.

yunhailuo:
> @yunhailuo black: https://black.readthedocs.io/

Thanks. Seems like a good tool to have and use.

ayan-b:
> I guess the question is what it is and if they are really official.

`black` claims [their styles](https://black.readthedocs.io/en/stable/the_black_code_style.html) are subset to PEP8.

xena_gdc_etl/xena_dataset.py
ayan-b: Hmm, seems a bug!

xena_gdc_etl/gdc.py
yunhailuo: You are right. Thanks.

xena_gdc_etl/xena_dataset.py
yunhailuo: Well. We can leave it that way. Just to say no need to spent time on `black`'s bug if not interesting...

xena_gdc_etl/main.py
yunhailuo: I guess it is since `black` is good. Just redo line breaks if you want.

xena_gdc_etl/gdc.py
ayan-b: Apparently, it throws E501.

ayan-b:
Now all the black lint fixes are in 866d14b.

yunhailuo:
Looks good. Really a lot of changes. Great job!

PR #38, opened 2019-05-31
Simplify travis tests

yunhailuo:
* Fix pandas, numpy and lxml version so that they are compatible with Python 3.7
* It seems stage only works with matrix expansion. Python matrix expansion only uses `python` and `env` as key. `MINIMUM_INSTALL` is thus used.


.travis.yml
```diff
@@ -1,46 +1,30 @@
...
-      script: make test
-    - python: 2.7
-      name: "python-2-min-req"
-      <<: *disable_global
-      install: make min-req-install
-      script: make test
+    - stage: lint
+      script: 
+        - pip install flake8
+        - flake8
```
ayan-b: Just to say that, since in this stage we are not disabling the `install` and `before_install`, they will be executed in travis.

yunhailuo: It seems fine but is indeed not necessary. Fix it in #37 if you want.

PR #40, opened 2019-06-06
test_gdc.py: Fix test_get_samples_clinical test

ayan-b:


ayan-b:
Understood, pushing a fix. :+1:

tests/test_gdc.py
```diff
@@ -84,7 +84,13 @@ def test_get_project_info():
...
     project_id = "TCGA-OV"
     actual = gdc.get_samples_clinical(project_id)
-    assert "2038fd65-d8f1-4b16-af90-b1c8f9a379a7" == actual['case_id'][0]
+    in_filter = {"project.project_id": project_id}
+    expected = gdc.search(
+        in_filter=in_filter,
+        fields="case_id",
+        endpoint="cases"
+    )
+    assert expected["case_id"][0] == actual["case_id"][0]
```
yunhailuo: Well, this feels like a "loop" test because `get_samples_clinica` is essentially defined by using `gdc.search`. But it's definitely much better and more stable. It should work as long as the return from GDC API is not randomly ordered. I'd say we merge this once tests pass. And if something goes wrong, we can change later.

ayan-b: Yup. I think the only other way we can achieving testing here by `patch`.

PR #41, opened 2019-06-12
Add retry functionality in download

ayan-b:


xena_gdc_etl/utils.py
```diff
@@ -178,3 +181,23 @@ def reduce_json_array(j):
     else:
         reduced = j
     return reduced
+
+
+def requests_retry_session(
+    retries=5,
+    backoff_factor=0.3,
+    status_forcelist=(500, 502, 504),
```
yunhailuo: Do we know if it fails on these three code? And just these three?

ayan-b:
> Have you tried this on one project with all data types (at least all genomics)?

Yes, it retries.

xena_gdc_etl/utils.py
ayan-b: So, we should disable it?

yunhailuo:
> > Have you tried this on one project with all data types (at least all genomics)?
> 
> Yes, it retries.

Interesting. It leaves trace in the log so you can tell?

xena_gdc_etl/utils.py
yunhailuo: What do you mean by "disable"?

ayan-b:
Yes, like this:
```
2019-06-12 18:35:58 [WARNING]: Retrying (Retry(total=2, connect=3, read=2$
2019-06-12 18:45:47 [WARNING]: Retrying (Retry(total=2, connect=3, read=2$
2019-06-12 18:48:55 [WARNING]: No methylation27 data for cohort TCGA-DLBC.
```

yunhailuo:
> Yes, like this:
> 
> ```
> 2019-06-12 18:35:58 [WARNING]: Retrying (Retry(total=2, connect=3, read=2$
> 2019-06-12 18:45:47 [WARNING]: Retrying (Retry(total=2, connect=3, read=2$
> 2019-06-12 18:48:55 [WARNING]: No methylation27 data for cohort TCGA-DLBC.
> ```

Oh. BTW, which log is this? Is it `etl.err`?

ayan-b:
Yes, it is `etl.err`.

xena_gdc_etl/utils.py
ayan-b: I mean it will retry regardless of error code.

yunhailuo: Sound too general. But I guess it doesn't hurt. Thanks.

PR #42, opened 2019-06-12
Use read_csv with sep="\t" instead of read_table

ayan-b:


xena_gdc_etl/main.py
```diff
@@ -159,7 +159,8 @@ def create_parser():
         required=True,
         help='A list of paths for Xena matrices files to be merged. All paths '
         'in this list support UNIX style pathname pattern expansion with '
-        '"glob". Files will be read by pandas.read_table.',
+        '"glob". Files will be read by pandas.read_table or pandas.read_csv '
+        'with sep="\t"',
```
yunhailuo: `read_table` is still in use?

ayan-b: Removed `read_table`.

PR #43, opened 2019-06-13
Log a unfinished.json file

ayan-b:
This will keep the dataypes and their corresponding
projects for which the ETL was not finished.

### DIFF
--- a/tests/test_gdc2xena.py
+++ b/tests/test_gdc2xena.py
@@ -8,6 +8,7 @@ from xena_gdc_etl import gdc2xena
 
 PATH = "tests/fixtures/"
 
+
 @pytest.mark.CI
 def test_logging():
     gdc2xena.gdc2xena(
--- a/tests/test_gdc2xena.py
+++ b/tests/test_gdc2xena.py
@@ -10,7 +10,7 @@ PATH = "tests/fixtures/"
 
 
 @pytest.mark.CI
-def test_logging():
+def test_unfinished_generation():
     gdc2xena.gdc2xena(
         root_dir=".",
         projects=["SOME_PROJECT1", "SOME_PROJECT2"],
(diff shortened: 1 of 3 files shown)


xena_gdc_etl/gdc2xena.py
```diff
@@ -84,6 +86,12 @@ def gdc2xena(root_dir, projects, xena_dtypes):
             try:
                 dataset.download().transform().metadata()
             except Exception:
+                if project not in unfinished:
+                    unfinished[project] = [dtype]
+                else:
+                    unfinished[project].append(dtype)
+                with open("unfinished.json", "w") as outfile:
+                    json.dump(unfinished, outfile)
```
yunhailuo: The file will be rewrite again every failed loop? And why not just print this summary at the end of log?

tests/test_gdc2xena.py
```diff
@@ -0,0 +1,24 @@
...
+import pytest
+
+from xena_gdc_etl import gdc2xena
+
+
+PATH = "tests/fixtures/"
+
+
+@pytest.mark.CI
+def test_logging():
```
yunhailuo: This is not real log if you don't put it to log. I'd suggest changing the name unless you do logging.

xena_gdc_etl/gdc2xena.py
ayan-b: This will allow us to check the unfinished datatypes on the fly.

ayan-b: So, that we can run another session on the instance downloading those datatypes.

yunhailuo: It's probably a bad idea for starting another session while the current session is still running and is already stressed server... But it's not a good solution anyway; so I'd say it's OK if this is good for you in real practice.

Another comment is where is this `unfinished.json`? Is it respecting the `root_dir` and other settings?

ayan-b: Yup, now it does.

PR #44, opened 2019-06-13
Improve logging

ayan-b:


ayan-b:
I think this PR is ready @yunhailuo.

yunhailuo:
The PR is good itself. But I'll wait until we get replies from GDC and figure out the "right" way to download from GDC. On hold for now.

PR #47, opened 2019-06-17
Add a CLI option to delete raw data upon generation of xena_matrix

ayan-b:
I am using this in the AWS. Basically, it deletes the raw data corresponding to a data type once the xena matrix is generated. Let me know if we need it, I will add tests and update the docs.

PR #48, opened 2019-06-19
 Skip the pipeline for a datatype if the Xena Matrix is found

ayan-b:
I am using this in the instance. Let me know if we need it, I will add tests and docs. This is on top of #47.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,7 +100,7 @@ Basic usage with command line tools
     mkdir -p /home/user/xena_root
     xge etl -P FM-AD -t htseq_counts htseq_fpkm htseq_fpkm-uq
 
-  If `-D` or `--delete` is specified the raw data corresponding to a datatype
+  If ``-D`` or ``--delete`` is specified the raw data corresponding to a datatype
   will be deleted upon generation of the xena matrix.
 
   Notes:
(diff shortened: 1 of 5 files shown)


yunhailuo:
In general, how do you know if the existing matrix is up to date and should be skipped? If you don't rely on some checks, you decide that arbitrarily. If we decide arbitrarily, we should just use a new script to be explicit about what exact project(s) and datatype(s) we want (like `--projects` and `--datatype`). Being implicit and just relying on filename check seems lazy and unstable.

ayan-b:
This command is kept for user convenience only and updating part is left to the user to check.

Suppose the pipeline is run and due to some errors (500 etc.), then either we can manually check or by `unfinished.json` and then execute the commands, but that would be somewhat more time-consuming.

> f we decide arbitrarily, we should just use a new script to be explicit about what exact project(s) and datatype(s) we want (like --projects and --datatype).

Avoided this since this the command will be really long.

yunhailuo:
Is good to have a check to see what matrix is there. But I really don't think filename check is the right way to go. And it's not worth the time you saved if you are not sure the data is right. I assume retry works, no 4xx response and you don't have too many unfinished to do?

ayan-b:
The ETL is done. Only setting the hub is left.

PR #50, opened 2019-06-21
Add GISTIC datatype

ayan-b:


xena_gdc_etl/resources/template.gistic.meta.json
```diff
@@ -0,0 +1,16 @@
+{"cohort":"{{ xena_cohort }}",
+ "url":"{% if gdc_release %}{{ gdc_release }}, {% endif %}https://api.gdc.cancer.gov/data/",
+ "dataSubType":"gene expression RNAseq",
```
yunhailuo: Need a new `dataSubType`. Probably need @maryjgoldman's input.

xena_gdc_etl/resources/template.gistic.meta.json
```diff
@@ -0,0 +1,16 @@
...
+ ":probeMap":"gencode.v22.annotation.gene.probeMap",
+ "dataProducer":"Genomic Data Commons",
+ "label":"{{ gdc_type }}",
+ "colNormalization":true,
+ "wrangling_procedure":
+ "Data from the same sample but from different vials/portions/analytes/aliquotes is averaged; data from different samples is combined into genomicMatrix; all data is then log2(x+1) transformed.",
```
yunhailuo: Again, need confirmation from @maryjgoldman because I'm not sure if log2 transformation is needed?

xena_gdc_etl/resources/template.gistic.meta.json
```diff
@@ -0,0 +1,16 @@
...
+ "PLATFORM":"Illumina",
+ "type":"genomicMatrix",
+ "wrangler":"Xena GDC ETL script",
+ "version":"{{ date }}",
+ "unit":"{% if unit %}log2({{ unit }}+1){% else %}log2(count+1){% endif %}"{% if notes %},
```
yunhailuo: Depends on the transformation above.

xena_gdc_etl/xena_dataset.py
```diff
@@ -366,6 +366,29 @@ def merge_sample_cols(
...
+    """Handles GISTIC Data type.
+
+    Args:
+        filelist (list of path): The list of input raw data.
+
+    Returns:
+        pandas.core.frame.DataFrame: Ready to load Xena matrix.
+    """
+
+    assert len(filelist) == 1
```
yunhailuo: Don't put it as a list then.

xena_gdc_etl/xena_dataset.py
```diff
@@ -366,6 +366,29 @@ def merge_sample_cols(
...
+    print('\rProcessing 1/1 file...', end='')
```
yunhailuo: Can be specific about which file is this since there is only and always one.

xena_gdc_etl/xena_dataset.py
```diff
@@ -932,6 +960,9 @@ class GDCOmicset(XenaDataset):
             snv_maf_matrix,
         )
     )
+    _RAWS2MATRIX_FUNCS['gistic'] = functools.partial(
+        handle_gistic,
+    )
```
yunhailuo: It's not partial so no need to be partial...

xena_gdc_etl/xena_dataset.py
ayan-b: Tried to follow: https://github.com/yunhailuo/xena-GDC-ETL/blob/34218373cc5669888d8e7b83a95ffe0f84d0e791/xena_gdc_etl/xena_dataset.py#L252

yunhailuo: Well, that's probably a wrong thing there... No need to follow if that looks wrong. I'll put it as your decision; either way.

xena_gdc_etl/xena_dataset.py
yunhailuo: Your decision as mentioned above. Either way is OK.

xena_gdc_etl/xena_dataset.py
yunhailuo: Actually, your comment above is great and reminds me. Maybe reuse `snv_maf_matrix`, make it more general so that we have one function for single matrix transformation which can be used in more datatypes.

xena_gdc_etl/xena_dataset.py
ayan-b: The code passes the argument as a list. So I guess we need to use list.

xena_gdc_etl/xena_dataset.py
yunhailuo: Like you would have `comment`, `usecols`, `index_col_name` (use in `set_index` rather than `read_csv`), `drop_col_name`, `rename_col_map`... And maybe use ID mapping for both. Not sure if this will work out with a nice and clean logic (like calculation?)...

yunhailuo: Maybe pass calculation in as a (lambda) function, though I'm not yet sure this is a good coding strategy... Could be good because it makes this function not tied to specific datatypes...

xena_gdc_etl/xena_dataset.py
yunhailuo: Oh. `raws2matrix`. Forgot that. You are right. My bad.

xena_gdc_etl/xena_dataset.py
```diff
@@ -366,6 +366,29 @@ def merge_sample_cols(
...
+
+    Args:
+        filelist (list of path): The list of input raw data.
+
+    Returns:
+        pandas.core.frame.DataFrame: Ready to load Xena matrix.
+    """
+
+    assert len(filelist) == 1
+    print('\rProcessing file...{}'.format(filelist[0]), end='')
```
yunhailuo: Filename after ellipsis? Isn't this sentence strange or it's just me?

xena_gdc_etl/resources/template.gistic.meta.json
```diff
@@ -0,0 +1,17 @@
...
+ "notes":"{{ notes }}"{% endif %},
+ "GDC_Description": "https://docs.gdc.cancer.gov/Data/Bioinformatics_Pipelines/CNV_Pipeline/"
+}
```
yunhailuo: Now that I'm seeing it and finding that the metadata could be very similar to CNV. I'm wondering if this datatype should use the CNV template?

ayan-b: But we need to see if we need `dataSubType` etc.

xena_gdc_etl/xena_dataset.py
```diff
@@ -366,6 +366,36 @@ def merge_sample_cols(
...
+    df = pd.read_csv(
+        filelist[0],
+        sep="\t",
+        header=0,
+        comment='#',
+        index_col=0,
+    )
+    df = df.drop(["Gene ID", "Cytoband"], axis=1)
+    mapping = gdc.map_fields(
+        project="TCGA-CHOL",
```
ayan-b: @yunhailuo How do you want this `project` to be passed to the `handle_gistic` function? I can extract it from the `filelist` or pass it as an argument.

yunhailuo: The initial reason for a project ID is to avoid extra searching since Xena matrix is done one project at a time. But you are right. The logic here is not good for passing that kind of specific info. I guess an alternative is to make the ID conversion a really specific one which accepts a list of input IDs. I'm not sure if this will work though since it may make the query really big. I may need to try something.

xena_gdc_etl/resources/template.gistic.meta.json
```diff
@@ -0,0 +1,17 @@
+{"cohort":"{{ xena_cohort }}",
+ "url":"{% if gdc_release %}{{ gdc_release }}, {% endif %}https://api.gdc.cancer.gov/data/",
+ "dataSubType": "copy number (gene-level)",
+ ":probeMap":"gencode.v22.annotation.gene.probeMap",
+ "dataProducer":"Genomic Data Commons",
+ "label":"{{ gdc_type }}",
+ "colNormalization":true,
+ "wrangling_procedure":
+ "Data from the same sample but from different vials/portions/analytes/aliquotes is averaged; data from different samples is combined into genomicMatrix.",
```
yunhailuo: Just to check, you are sure the "Data from the same sample but from different vials/portions/analytes/aliquotes is averaged;", aren't you?

xena_gdc_etl/resources/template.gistic.meta.json
```diff
@@ -0,0 +1,17 @@
+{"cohort":"{{ xena_cohort }}",
+ "url":"{% if gdc_release %}{{ gdc_release }}, {% endif %}https://api.gdc.cancer.gov/data/",
+ "dataSubType": "copy number (gene-level)",
+ ":probeMap":"gencode.v22.annotation.gene.probeMap",
+ "dataProducer":"Genomic Data Commons",
+ "label":"{{ gdc_type }}",
+ "colNormalization":true,
```
yunhailuo: Are you sure?

xena_gdc_etl/utils.py
```diff
@@ -201,3 +201,40 @@ def requests_retry_session(
...
+        list_1 (list): First list.
+        list_2 (list): Second list.
+
+    Returns:
+        list: The intersection of the two lists.
+    """
+
+    set_1 = set(list_1)
+    set_2 = set(list_2)
+    return list(set_1.intersection(set_2))
```
yunhailuo: I think this functionality is under development in general. But here, I don't understand why a one-line code becomes a new separated function?

xena_gdc_etl/xena_dataset.py
yunhailuo: Maybe GraphQL can do a better and cleaner job here...

xena_gdc_etl/utils.py
ayan-b: Actually changes the logic here now. I guess now it deserves to be a new function?

xena_gdc_etl/resources/template.gistic.meta.json
ayan-b: Not really, but to quote from GDC:
```
The Copy Number Liftover Workflow uses the TCGA level 2 tangent.copynumber
files described above. These files were generated by first normalizing array intensity
values, estimating raw copy number, and performing tangent normalization, which
subtracts variation that is found in a set of normal samples.
```

xena_gdc_etl/resources/template.gistic.meta.json
ayan-b: Not really, but again to quote from GDC:
```
A table that associates contiguous chromosomal segments with genomic coordinates,
mean array intensity, and the number of probes that bind to each segment.
```

xena_gdc_etl/resources/template.gistic.meta.json
ayan-b: Reusing now, but not really sure if it is an elegant solution ...

xena_gdc_etl/xena_dataset.py
ayan-b: But still, the query will be big for all the project.

xena_gdc_etl/resources/base_cnv.json.jj2
```diff
@@ -0,0 +1,8 @@
+ "cohort":"{{ xena_cohort }}",
+ "url":"{% if gdc_release %}{{ gdc_release }}, {% endif %}https://api.gdc.cancer.gov/data/",
+ "dataProducer":"Genomic Data Commons",
+ "label":"{{ gdc_type }}",
+ "wrangler":"Xena GDC ETL script",
+ "version":"{{ date }}",{% if notes %}
+ "notes":"{{ notes }}",{% endif %}
+ "GDC_Description": "https://docs.gdc.cancer.gov/Data/Bioinformatics_Pipelines/CNV_Pipeline/"
```
yunhailuo: Not sure if this is beneficial. For one thing it makes another layer in terms of data structure while the gain is not quite clear to me. For another thing, you can, in principle, make a base for all datatypes (like just remove `GDC_Description`). Just have it for CNV is like half done. I'd say go back to the flat structure with each datatype has their own template.

ayan-b: Yes, I also don't think it is a good solution.

xena_gdc_etl/resources/template.gistic.meta.json
yunhailuo: No. It's not. You are definitely right. The two are different enough to have different templates.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +493,45 @@ def gdc_check_new(new_file_uuids):
...
+    field_2_last = fields_2[-1]
+    for hit in result:
+        for field in field_intersections:
+            hit = hit[field]
+        for data in hit:
+            field_1_leaves = []
+            field_2_leaves = []
+            extract_leaves(data, field_1_last, field_1_leaves)
+            extract_leaves(data, field_2_last, field_2_leaves)
+            assert len(field_2_leaves) == 1
```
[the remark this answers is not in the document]
ayan-b: This confusing to me as well, consider the following mapping: A -> a, A -> b, B -> c, C -> d. Now, what should be considered as unique? Capital letters or small letters?

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +493,45 @@ def gdc_check_new(new_file_uuids):
...
+    mapping = {}
+    fields_1 = field_1.split(".")
+    fields_2 = field_2.split(".")
+    field_intersections = get_intersection(fields_1, fields_2)
+    field_1_last = fields_1[-1]
+    field_2_last = fields_2[-1]
+    for hit in result:
+        for field in field_intersections:
+            hit = hit[field]
+        for data in hit:
```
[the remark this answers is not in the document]
ayan-b: Yes. `hit` seems to be always a list even if the number of elements in one.

ayan-b:
> Which field is unique or it's never unique?

It should work even if it is not unique. In that case we need to remove the check `assert len(field_2) == 1`. If it is not unique, the mapping would be list to list.

> Certain nodes can be dict or list but it's not predictable in general.

We can take help of `last`.

xena_gdc_etl/utils.py
```diff
@@ -201,3 +201,44 @@ def requests_retry_session(
...
+            are to be collected.
+        last (str): If a key matches with ``last``, corresponing value is
+            appended to leaves.
+        leaves (list): Used to store the leaves.
+    """
+
+    if isinstance(data, dict):
+        for key, value in data.items():
+            if key == last:
+                leaves.append(value)
```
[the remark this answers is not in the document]
ayan-b: I wonder then how the correct key would be identified. For the given structure how `submiiter_id`/`aliquot_id` will be identified since the `value` for both of them will be a `str`(neither `dict` nor `list`).

xena_gdc_etl/gdc.py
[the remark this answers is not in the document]
yunhailuo: What will be the first `hit` if I have `endpoint='cases', project='TCGA-CHOL', field_1='project.name', field_2='project.project_id'`?

xena_gdc_etl/gdc.py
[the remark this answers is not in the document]
yunhailuo: Confusing indeed. I brought us to a completely wrong direction in the first place... One of many reasons I decided to give up on this "too-general" challenge.

xena_gdc_etl/resources/template.gistic.meta.json
yunhailuo: Please check Mary's response on slack related to this.

xena_gdc_etl/utils.py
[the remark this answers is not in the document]
yunhailuo: Oh. Sorry. I missed it. Thought you have the field path here.

xena_gdc_etl/gdc.py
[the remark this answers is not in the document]
ayan-b: Agreed, not a `list` here.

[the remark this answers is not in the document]
yunhailuo: Another indication suggesting I'm wrong about this direction. Really confusing and difficult to find these potential exceptions. My bad.

yunhailuo:
> It should work even if it is not unique. In that case we need to remove the check `assert len(field_2) == 1`. If it is not unique, the mapping would be tuple to list (not implemented in the code).

Not ideal and confusing because the return of this function become "unpredictable" and may requires some understanding of details in this function. To me, such a non-black-box function is not a good function...

> It seems all the nodes are `list`s except for the leaves. The recursion should take care of them as long as they are either `dict` or `list`.

I think being unpredictable is bad enough. You already missed a simple exception above...

> If it does not overlap, there shouldn't be any problem considering the `mapping` will be inside each hit.

I'm not sure about that. Wrong map could still happen within a hit. Say a hit is one sample, one portion id could be mapped to another aliquot (don't think it will happen for this case; just an abstracted example).

yunhailuo:
Once we get the submitter_id of aliquot, we can think about trimming it down to sample submitter_id.

xena_gdc_etl/xena_dataset.py
ayan-b: Moved to #52 .

xena_gdc_etl/gdc.py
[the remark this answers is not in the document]
ayan-b: Moved to #52.

ayan-b:
@yunhailuo  Updated the code, please review.

xena_gdc_etl/xena_dataset.py
ayan-b: I was wondering how can we pass something like `df['dna_vaf'] = df['t_alt_count'] / df['t_depth']` ...

yunhailuo:
One related question and one unrelated question:

* You've tried with the exact code b00d5bc866e3d4cc161bc663d67dae25fb46a069 and knew that it can be load to Xena correctly, right?
* Do you remember why we have and use this: https://github.com/yunhailuo/xena-GDC-ETL/blob/master/tests/utils.py ?

yunhailuo:
Please remove `colNormalization` as Mary suggested. No need for another test if you've tested b00d5bc866e3d4cc161bc663d67dae25fb46a069.

ayan-b:
> You've tried with the exact code b00d5bc and knew that it can be load to Xena correctly, right?

Yes.

> Do you remember why we have and use this: https://github.com/yunhailuo/xena-GDC-ETL/blob/master/tests/utils.py ?

To compare dictionaries, for example here:
https://github.com/yunhailuo/xena-GDC-ETL/blob/31db69c796f8c55e54857ae791e9f6da3b420e56/tests/test_gdc.py#L29

yunhailuo:
Thanks. PR is ready to go.

Why we need `compare_dict(expected, actual)` instead of `expected == actual`? Can you give one specific example showing the two are different?

ayan-b:
In python dict order is not preserved, so you may get AssertionError if they are asserted like that. I don't know why but this happens mostly in python 2.7.

PS: Looking for the travis job it failed.

yunhailuo:
> In python dict order is not preserved, so you may get AssertionError if they are asserted like that. I don't know why but this happens mostly in python 2.7.
> 
> PS: Looking for the travis job it failed.

Thanks. It's possible python 2 and 3 are different. But please help me find an example so that we are sure this is necessary.

ayan-b:
In this build 3.4 & 2.7 failed https://travis-ci.com/yunhailuo/xena-GDC-ETL/builds/103860400.

yunhailuo:
> In this build 3.4 & 2.7 failed https://travis-ci.com/yunhailuo/xena-GDC-ETL/builds/103860400.

Thanks. It actually helps me find out a set of bugs which I should have noticed before (sorry). And `compare_dict` doesn't really help with the failure you showed. We can close the discussion here and move to #54

PR #52, opened 2019-06-26
Add mapping functionality

ayan-b:


ayan-b:
@yunhailuo Please review.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,60 @@ def gdc_check_new(new_file_uuids):
...
+    mapping = {}
+    field_intersections = object_path.split(".")
+    fields_1 = field_1.split(".")
+    fields_2 = field_2.split(".")
+    field_1_last = fields_1[-1]
+    field_2_last = fields_2[-1]
+    for hit in result:
+        for field in field_intersections:
+            if isinstance(hit, list):
+                hit = hit[0]
```
yunhailuo: Get first hit will give you failure like this: https://github.com/yunhailuo/xena-GDC-ETL/blob/yl/test-map/tests/test_gdc.py#L160-L188

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,60 @@ def gdc_check_new(new_file_uuids):
...
+    input_id_name,
+    output_id_name,
+):
+    """Maps two fields in a dictionary.
+    Args:
+        endpoint (str): One string of GDC API supported endpoint.
+        object_path:
+        input_ids:
+        input_id_name:
+        output_id_name:
```
yunhailuo: Unfinished?

ayan-b:
For property mapping within the same object, the problem is how to reach that subtree. There is no common trend if the intermediate nodes are `list`s or `dict`s.

yunhailuo:
> For property mapping within the same object, the problem is how to reach that subtree. There is no common trend if the intermediate nodes are `list`s or `dict`s.

Not sure what you mean by "reach that subtree". My imagination is `object_path` is the path leading to that same object. Both input and output properties are guaranteed to be both in that object. So once you get there, you only need to get two properties and put them in the return map.

yunhailuo:
I might be able to make a prototype tomorrow if that helps explain what I mean.

Don't get me wrong though. It will be extremely useful if you have a nice and clean, thus stable, solution. I just worried if that's too difficult to worth the effort.

ayan-b:
@yunhailuo Please review.

ayan-b:
@yunhailuo Failing because `last` of both of them is `submitter_id`. Need to fix that.

yunhailuo:
> @yunhailuo Failing because `last` of both of them is `submitter_id`. Need to fix that.

Right. That's on purpose. It happens when the two properties are not in the same object.

ayan-b:
It's hard to fix, to be honest. :sweat_smile:

yunhailuo:
> It's hard to fix, to be honest. 😅

I know. That's why I'm not convinced this will work and suggest to have a more feasible problem to tackle and focus on same object and two bare properties. But still your call if you still want to tackle this. My only request is the solution must be stable if we go that way.

ayan-b:
> tackle and focus on same object and two bare properties

Should work for the present code I believe.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,96 @@ def gdc_check_new(new_file_uuids):
...
+    result = search(
+        endpoint=endpoint,
+        in_filter=in_filter,
+        fields=[field_1, field_2],
+        typ="json",
+    )
+    mapping = {}
+    field_intersections = object_path.split(".")
+    fields_1 = field_1.split(".")
+    fields_2 = field_2.split(".")
```
yunhailuo: I guess I'm confused. I saw here and thought you still want to tackle the harder problem and don't want to just focus on same project and two bare properties.

ayan-b: Removing.

ayan-b:
@yunhailuo Please review.

yunhailuo:
OK. I'd like a clarification/specification here. What is the problem you are trying to workout here, general one, specific one (same object with two bare properties), or something else?

ayan-b:
You can say general one minus `last` having the same name. :sweat_smile:

yunhailuo:
> You can say general one minus `last` having the same name. 😅

OK. To me a good function is
1. Has a clear and simple to get definition about what it does and what it doesn't (which handle errors properly and not return unexpected results)
2. As blackbox as possible.

So can you layout what this `map_fields`  does like you have full spec doc for it?

ayan-b:
Even if I deal with the specific one, the code won't be much different I guess.

yunhailuo:
Let me say it this way. I'm not going to throw out a bunch of code and just say it does this and might do others but not guaranteed. As a users, I don't want to use that code because I'm either forced to understand details of its implementation or suffered one day from its potential instability. As a reviewer, I don't know what to test, especially those potential but not guaranteed functions. So I'm going to be strict on this and do it by
* First a clear and simple to understand definition;
* Second a nice and clean logic to address the definition above;
* Third review implementation;
* Four test to make sure if stable.

ayan-b:
Sorry if I was not clear before. This function only deals with `specific one` . The approach for `specific one` is same as `general one` - input and output ids having the same `last`.

yunhailuo:
> Sorry if I was not clear before. This function attempts deals with `specific one` . The approach for `specific one` is same as `general one` - input and output ids having the same `last`.

That's a good start. Let's define it as you are saying to another new user or like you will put in the doc. People wouldn't understand what is `specific one`, would they?

yunhailuo:
Or if you want, just update the doc part of the code and let's review that first.

ayan-b:
Updated docstrings.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,111 @@ def gdc_check_new(new_file_uuids):
...
+    endpoint,
+    object_path,
+    input_ids,
+    input_id_name,
+    output_id_name,
+):
+    """Maps two fields - ``input_id_name`` and ``output_id_name`` from GDC API
+    in a dictionary. The fields should be within same ``object`` defined by
+    the argument ``object_path``. It is assumed that each input_id has only
+    one corresponding output_id.
```
yunhailuo: For a single object, JSON schema doesn't allow two properties having the same name. No need to assume.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,111 @@ def gdc_check_new(new_file_uuids):
...
+
+def map_fields(
+    endpoint,
+    object_path,
+    input_ids,
+    input_id_name,
+    output_id_name,
+):
+    """Maps two fields - ``input_id_name`` and ``output_id_name`` from GDC API
+    in a dictionary. The fields should be within same ``object`` defined by
```
yunhailuo: object is one arg? Wondering why it's in code style.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,111 @@ def gdc_check_new(new_file_uuids):
...
+    the argument ``object_path``. It is assumed that each input_id has only
+    one corresponding output_id.
+
+    Args:
+        endpoint (str): One string of GDC API supported endpoint.
+        object_path (str): Path leading to that same object (input_id_name
+            and output_id_name).
```
yunhailuo: Maybe a rewording here. Here, it's just a path to one object. Then down below, `input_id_name` and `output_id_name` are properties of that object.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,111 @@ def gdc_check_new(new_file_uuids):
...
+        input_ids (list of str or str): A list of input ids which is to mapped.
+        input_id_name (str): The id which is to be mapped to
+            ``output_id_name``.
+        output_id_name (str): ``input_id_name`` is to be mapped with
+            ``output_id_name``.
```
yunhailuo: You want to restrict all these to just IDs?

yunhailuo: Either way, give some guidance (examples, links, etc) on what is the correct name/ID for your function.

xena_gdc_etl/gdc.py
yunhailuo: I mean direct properties.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,125 @@ def gdc_check_new(new_file_uuids):
...
+        object_path: "samples.portions.analytes.aliquots"
+        input_id_name: "aliquot_id"
+        output_id_name: "submitter_id"
+
+    However the following is not a valid one:
+        object_path: "samples.portions"
+        input_id_name: "analytes.aliquots.submitter_id"
+        output_id_name: "submitter_id"
+    Note that it is invalid because ``input_id_name`` is nested i.e., it has
+    two dots.
```
yunhailuo: Should be fine. I think there is an advanced way to show example in docstring?

yunhailuo: There are some examples out there but it might be too much for here...

xena_gdc_etl/gdc.py
ayan-b: I meant something like this is not allowed https://api.gdc.cancer.gov/cases/?fields=samples.portions.analytes.aliquots.aliquot_id,samples.submitter_id&filters={%22op%22%3A%22and%22%2C%22content%22%3A[{%22op%22%3A%22in%22%2C%22content%22%3A{%22field%22%3A%22samples.submitter_id%22%2C%22value%22%3A[%22TCGA-4G-AAZO-11A%22]}}]} since TCGA-4G-AAZO-11A has 3 `aliquot_id`s.

xena_gdc_etl/gdc.py
ayan-b: Changed.

xena_gdc_etl/gdc.py
ayan-b: Examples are now inside `Examples:`.

xena_gdc_etl/gdc.py
ayan-b: Updated the docstring.

xena_gdc_etl/gdc.py
yunhailuo: I'm actually wondering if this should be split to individual argument as explaining what is a valid `object_path`, what is a valid `input_id_name` and what is a valid `output_id_name`. To be exact (again being explicit and precision is important):
* `object_path`: is `samples.portions.analytes.aliquots.aliquot_id` valid? If not, you need to clarify here to help users and you need to handle that error in the code later on;
* `input_id_name`: since its values will be used in 1) GDC query filter; 2) key in return map. Technically, it can be anything whose value a) can help return a non-empty result from GDC; b) hashable so that can be key in dict. Again, some if not all needs explanation to users and some if not all needs error handling with helpful message.
* `output_id_name`: it is only used for getting value for the return map. Technically, it can be anything even it's
  * Not in the object: just return `None` or `''`;
  * Property pointing to another object nested in this object: just return the value (list or dict);

No error handling is needed for `output_id_name` case since there is no technical errors. But users need to know this behavior in doc rather than reading the code. Actually, it's said in [Google's python style guide](http://google.github.io/styleguide/pyguide.html#383-functions-and-methods) which I found useful:
> A docstring should give enough information to write a call to the function without reading the function’s code.

ayan-b: Changed id to field in order to reflect that they can also be non-ids.

xena_gdc_etl/gdc.py
ayan-b: Examples are already down below, wondering if it is needed here ...

xena_gdc_etl/gdc.py
yunhailuo: Let me understand this: TCGA-4G-AAZO-11A is the ID for a sample object and aliquot_ids is the ID for an aliquot object which is nested in the TCGA-4G-AAZO-11A object. Are you doing two properties of the same object or not? I guess that's why a clear definition is important at least to me.

yunhailuo:
I'll let you edit it and review later.

xena_gdc_etl/gdc.py
ayan-b: > input_id_name: since its values will be used in 1) GDC query filter; 2) key in return map. Technically, it can be anything whose value a) can help return a non-empty result from GDC; b) hashable so that can be key in dict. Again, some if not all needs explanation to users and some if not all needs error handling with helpful message.

I think it should only be a `str` if it is a tuple of string that can get us results from API but won't be able to map.

xena_gdc_etl/gdc.py
ayan-b: Using lists now.

xena_gdc_etl/gdc.py
yunhailuo: Well, they should all three be string. I was trying to say aside from being string what string it can be...

And the hashable part is for its value which should be in `input_field_items`. It can be a number, for example, if we decide input is field rather than just ID.

ayan-b:
Addressed most of the comments.

yunhailuo:
> Addressed most of the comments except for `aside from being string what string it can be` which needs more investigating.

OK. Will mark all previous comments as resolved and do a new review.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,154 @@ def gdc_check_new(new_file_uuids):
...
+    object_path,
+    input_field_items,
+    input_field_name,
+    output_field_name,
+):
+    """Maps two fields - ``input_field_name`` and ``output_field_name`` from
+    the GDC API in a dictionary. The fields should be within same object
+    defined by the argument ``object_path``. The ``input_field_name`` and
+    ``output_field_name`` should not be nested. The last property shouldn't be
+    same for ``input_field_name`` and ``output_field_name``.
```
yunhailuo: Wondering what will happen if the "last property" are the same. We will see when you get to implementation.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,154 @@ def gdc_check_new(new_file_uuids):
...
+        It is invalid since the last property ("submitter_id") is same for
+        both ``input_field_name`` and ``output_field_name``.
+
+    Args:
+        endpoint (str): One string of GDC API supported endpoint.
+        object_path (str): Path leading to the same object (input_field_name
+            and output_field_name). It should point to an object and should
+            not be a property itself. For example,
+            "samples.portions.analytes.aliquots.aliquot_id" is an invalid
+            ``object_path`` but "samples.portions.analytes.aliquots" is not.
```
yunhailuo: > is not
is right in logic but users may need a second thought before realizing "samples.portions.analytes.aliquots" is correct. I'd suggest to make it more simple and clear.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,154 @@ def gdc_check_new(new_file_uuids):
...
+        input_field_items (list of str or str): A list of items that are of
+            input field type which are to mapped. The items should be hashable
+            since they will be used as key in the returned map.
+        input_field_name (str): A direct property of the object pointed by
+            ``object_path`` the field which is to be mapped to
+            ``output_field_name``. List of all available fields can be found
+            here https://api.gdc.cancer.gov/files/_mapping.
```
yunhailuo: Why restrict it to the "files" endpoint? Is this a good one: https://docs.gdc.cancer.gov/API/Users_Guide/Appendix_A_Available_Fields/ ?

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,154 @@ def gdc_check_new(new_file_uuids):
...
+
+    Examples:
+        Vaild examples would be:
+        1.  object_path: "samples.portions.analytes.aliquots"
+            input_field_name: "aliquot_id"
+            output_field_name: "submitter_id"
+
+        2.  object_path: "samples.portions"
+            input_field_name: "analytes.aliquots.submitter_id"
+            output_field_name: "submitter_id"
```
yunhailuo: Can you translate this into a test case in `test_map_fields`? That may help me understand better what you want to do and why you say
> The last property shouldn't be same for ``input_field_name`` and ``output_field_name``.

ayan-b:
> Layout your algorithm in simple pseudo code if possible

The code follows https://en.wikipedia.org/wiki/Depth-first_search.

yunhailuo:
> > Layout your algorithm in simple pseudo code if possible
> 
> The code follows https://en.wikipedia.org/wiki/Depth-first_search.

OK. I'm going to review the code then.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,156 @@ def gdc_check_new(new_file_uuids):
...
+            elif isinstance(hit, dict):
+                if object_path[index] in hit:
+                    hit = hit[object_path[index]]
+        if not isinstance(hit, list):
+            hit = [hit]
+        for data in hit:
+            field_1_leaves = []
+            field_2_leaves = []
+            extract_leaves(data, input_field_last, field_1_leaves)
+            extract_leaves(data, output_field_last, field_2_leaves)
```
yunhailuo: Data in the hit here is the object pointed by `object_path`, isn't?

Not knowing what your test case I asked in the previous comment would be, I assume the next step is just to get input field and output field from this object `data`. Why it's not just `data.get(input_field_last)` and `data.get(output_field_last)` here?

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,156 @@ def gdc_check_new(new_file_uuids):
...
+        for index in range(len(object_path)):
+            if isinstance(hit, list):
+                get_mapping(
+                    hit,
+                    object_path[index:],
+                    mapping,
+                    input_field_last,
+                    output_field_last,
+                    input_field_items,
+                )
```
yunhailuo: Recursion has a default depth in python and recursion has overhead for every call. I think it's OK to use recursion here, assuming we won't have deep nesting or performance hurdle. But just so you know, with a defined object path, it's possible and easy to do it in an iterative way. This [article](https://towardsdatascience.com/how-to-flatten-deeply-nested-json-objects-in-non-recursive-elegant-python-55f96533103d) compares the two and listed some advantages.

yunhailuo:
Two comments for now just to straight out the algorithm. We will need to have some error handling later. Please remember to layout the test case so that I know your expectation for those scenario.

xena_gdc_etl/gdc.py
ayan-b: I wonder if it will work for field_names having dots like "portions.analytes.aliquots.aliquot_id" with object path "samples". As we have not yet reached to the `aliquots` point via object_path.

yunhailuo: That's why I repetitively and annoyingly force you to do definition and why I so want to see your test case. I don't really understand why you still have those property names as valid while saying you want to do two properties from the same objects... What I do know is no object will have property name like "portions.analytes.aliquots.aliquot_id". To be clear, I'm still not 100% sure what you expect to accomplish with this `map_fields` function. Let's put aside the definition and focus on what you expect to see with your valid example #2.

xena_gdc_etl/gdc.py
ayan-b: Updated the link.

xena_gdc_etl/gdc.py
ayan-b: Updated.

xena_gdc_etl/gdc.py
yunhailuo: I have to say I fail to make myself clear either. Sorry about that. Please feel free to check GDC's schema and get a feeling about what is an object, what is their "nesting" mechanism and how the schema works: https://raw.githubusercontent.com/NCI-GDC/portal-ui/develop/data/schema.json

ayan-b: Is this a valid call?
```python
gdc.map_fields(
        endpoint="cases",
        object_path="samples",
        input_field_items=[
            "35f8d837-f78c-4f88-a4cd-50ea2f9f9437",
            "48de3d5a-35d8-456d-a23e-e2dc25a840ac",
            ],
        input_field_name="portions.analytes.aliquots.aliquot_id",
        output_field_name="submitter_id",
   )
````

Comment (yunhailuo) [reply xena_gdc_etl/gdc.py]: What's your expected result for this one?

Comment (ayan-b) [reply xena_gdc_etl/gdc.py]:
```python
{
     "35f8d837-f78c-4f88-a4cd-50ea2f9f9437": "TCGA-4G-AAZO-11A",
     "48de3d5a-35d8-456d-a23e-e2dc25a840ac": "TCGA-4G-AAZO-11A",
}
```

yunhailuo: OK. Get it. Please ignore all my previous suggestion and push a code to do that and we move forward from there.

ayan-b: Added the test case.

xena_gdc_etl/gdc.py
ayan-b: Fixed that. No issue now if they are the same. :tada:

xena_gdc_etl/gdc.py
ayan-b: Sorry, this is no longer valid. They can be the same. Added a test case too: https://github.com/yunhailuo/xena-GDC-ETL/pull/52/files#diff-063b435e0ffe76e782c7dd8dd3aebabaR163

ayan-b:
OK, I realize now my understanding of `object_path` was wrong. I thought it is till "that point" we will traverse so that we have a one to one mapping which is clearly wrong. This PR aims to add that feature considering the definition of `object_map`. The code should work for all possible arrangements of `object_map`, `input_field_name` and `output_field_name`, in a way it can also be thought of as an extension of the "general case" you defined earlier.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,147 @@ def gdc_check_new(new_file_uuids):
...
+            ``output_field_name``. Keys are ``input_field_items`` and the
+            corresponding values are mapped ``output_field_items``.
+    """
+
+    if isinstance(input_field_items, str):
+        input_field_items = [input_field_items]
+    field_1 = object_path + "." + input_field_name
+    field_2 = object_path + "." + output_field_name
+    in_filter = {}
+    in_filter[field_1] = input_field_items
```
yunhailuo: Why this dict is defined this way? Not `in_filter = {field_1: input_field_items}`?

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,147 @@ def gdc_check_new(new_file_uuids):
...
+            extract_leaves(data, output_field_name, field_2_leaves)
+            for leaf in field_1_leaves:
+                if leaf in input_field_items:
+                    if (
+                        isinstance(field_2_leaves, list)
+                        and len(field_2_leaves) == 1
+                    ):
+                        mapping[leaf] = field_2_leaves[0]
+                    else:
+                        mapping[leaf] = field_2_leaves
```
yunhailuo: I'd personally prefer stable return meaning values are always list.

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,147 @@ def gdc_check_new(new_file_uuids):
...
+    Returns:
+        dict: The mapping between ``input_field_items`` and
+            ``output_field_name``. Keys are ``input_field_items`` and the
+            corresponding values are mapped ``output_field_items``.
+    """
+
+    if isinstance(input_field_items, str):
+        input_field_items = [input_field_items]
+    field_1 = object_path + "." + input_field_name
+    field_2 = object_path + "." + output_field_name
```
yunhailuo: If you end up want to do it for general use, object_path is not really needed since you can derive it as you did initially.

xena_gdc_etl/utils.py
```diff
@@ -201,3 +201,31 @@ def requests_retry_session(
...
+            else:
+                extract_leaves(
+                    value,
+                    field_name,
+                    leaves,
+                    current_field + key + ".",
+                )
+    elif isinstance(data, list):
+        for _, x in enumerate(data):
+            extract_leaves(x, field_name, leaves, current_field)
```
yunhailuo: What's the point for enumeration instead of looping?

xena_gdc_etl/gdc.py
```diff
@@ -488,3 +492,147 @@ def gdc_check_new(new_file_uuids):
...
+                    object_path[index:],
+                    mapping,
+                    input_field_name,
+                    output_field_name,
+                    input_field_items,
+                )
+                return
+            elif isinstance(hit, dict):
+                if object_path[index] in hit:
+                    hit = hit[object_path[index]]
```
yunhailuo: What if it's not in `hit`? Should it be for sure in `hit`, assuming GDC API does things correctly (meaning we don't have to handle GDC's bug).

ayan-b:
Addressed all of the comments. Now that `object_path` definition is clear to me lets close this PR and start afresh the specific case in another PR.

yunhailuo:
Had a try in #53 . Very crude and preliminary. Hopefully tests pass and there isn't too many bugs...

Have to go and will check back later.

PR #53, opened 2019-06-28
Try to map two fields from GDC data

yunhailuo:


ayan-b:
@yunhailuo I tried this code in https://github.com/ayan-b/xena-GDC-ETL/tree/gistic2, but for larger projects, I am getting Response 414 (Request-URI too large). So I think we can split the queries in a certain number of segments so that the URI length is under the limit.

yunhailuo:
Apparently a bug. I knew there must be some! Fixed in #9a0b488

yunhailuo:
> @yunhailuo I tried this code in https://github.com/ayan-b/xena-GDC-ETL/tree/gistic2, but for larger projects, I am hitting `Response 414` (Request-URI too large). So I think we can split the queries in a certain number of segments so that the URI length is under the limit or we can use `POST` request [here](https://github.com/yunhailuo/xena-GDC-ETL/pull/53/files#diff-128b8a60ab6a7072f3ae3d141556dc15R522) as [suggested by GDC](https://docs.gdc.cancer.gov/API/Users_Guide/Search_and_Retrieval/#components-of-a-request) for larger queries.

Apparently a bug. I knew there must be some! Fixed in @9a0b488

ayan-b:
Tried with `TCGA-BRCA`. Looks good. Merge this now?

yunhailuo:
Give me a day. I want to try more corner cases.

yunhailuo:
Found and fixed another bug. I worried there might be more. But it's clear we should move forward. Merge this now.

Issue #61, opened 2019-07-01
Add min:-1 max:1 to GISTIC .json files 


maryjgoldman:
Add min:-1 max:1 to the .json files that go with the GISTIC files. This will help the data display better.

This will look similar to how the DNA methylation data has min:0 max:1

maryjgoldman:
This looks good for the CHOL dataset on the hub. Please update this for all other GISTIC datasets in all other cohorts.

ayan-b:
Added for all cohorts.

maryjgoldman:
QA for amplified genes looks good. Attaching results from my tests. Still need to test deleted genes.

Example 1:
![Screen Shot 2019-07-02 at 2 51 45 PM](https://user-images.githubusercontent.com/13877333/60549214-12dfbd80-9cd9-11e9-9ed9-6fcc301317d7.png)
[GBM_CDKN2A_CDKN2B.txt](https://github.com/ucscXena/xena-GDC-ETL/files/3352447/GBM_CDKN2A_CDKN2B.txt)

Example 2:
![Screen Shot 2019-07-02 at 2 55 23 PM](https://user-images.githubusercontent.com/13877333/60549376-8b467e80-9cd9-11e9-8656-38521fb3a3b3.png)
[BRCA_ERBB2_Her2_immuno_score.txt](https://github.com/ucscXena/xena-GDC-ETL/files/3352456/BRCA_ERBB2_Her2_immuno_score.txt)

maryjgoldman:
Looking at genes that are typically deleted, I do not see a close correspondence between the segmented CNV and the GISTIC data. While this is worrisome, I see the same trend in the legacy TCGA data between the GISTIC data generated by Firehose and the segmented data generated by TCGA. So perhaps this is a well known trend.

Looked at genes in https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2921682/

New GDC data:
![Screen Shot 2019-07-03 at 9 28 21 AM](https://user-images.githubusercontent.com/13877333/60608902-5ab42280-9d75-11e9-88bd-e60e16f0a551.png)
Bookmark: [GDC_LUAD_FHIT.txt](https://github.com/ucscXena/xena-GDC-ETL/files/3355919/GDC_LUAD_FHIT.txt)

Legacy TCGA:
[...]

PR #65, opened 2019-07-02
template.gistic.meta.json: Add min and max

ayan-b:
Closes https://github.com/yunhailuo/xena-GDC-ETL/issues/61


yunhailuo:
Can we try one matrix with those changes (meaning doesn't have to be from dece5370552ddc317758aa5ad09e68766d4983ff). I'm wondering if those digits matters and would like to be sure it works as @maryjgoldman expected. The patch seems quite right.

ayan-b:
Added for TCGA-CHOL in the hub.

maryjgoldman:
So to look at this I should look at the hub for the GISTIC CHOL dataset?

yunhailuo:
> So to look at this I should look at the hub for the GISTIC CHOL dataset?

I guess so, @ayan-b ?

ayan-b:
Yes.

PR #67, opened 2019-07-02
Filter for blood derived samples in CNV data

### LINKED ISSUE #60: Remove blood normals from new segmented copy number datasets
The new segmented copy number datasets (DNAcopy) have copy number data for samples that are blood normal. We only want copy number data for the tumor samples. 

Screen Shot 2019-07-01 at 4 16 42 PM

The samples in black are the samples for which we want to remove the segmented copy number data for.

[newCNV-bloodnormaltoberemoved.txt](https://github.com/yunhailuo/xena-GDC-ETL/files/3347628/newCNV-bloodnormaltoberemoved.txt)

This is a bookmark of the above data. It can be imported back into Xena via the bookmark menu

ayan-b:
Closes #60.


ayan-b:
Updated CNV data into the hub.

PR #68, opened 2019-07-02
Remove phenotype data having no genomic data

### LINKED ISSUE #63: Samples with phenotype data but no genomic data are still being included
These are samples that have slide image data and phenotype data but no genomic data. These samples tend to end in 'Z', rather than 'A' or 'B'

Screen Shot 2019-07-01 at 4 25 04 PM
Samples that need to be excluded are in a black rectangle

[tooManySamples.txt](https://github.com/yunhailuo/xena-GDC-ETL/files/3347644/tooManySamples.txt)
Bookmark to see this for BRCA

ayan-b:
Closes #63

### DIFF
--- a/xena_gdc_etl/xena_dataset.py
+++ b/xena_gdc_etl/xena_dataset.py
@@ -1289,7 +1289,6 @@ class GDCPhenoset(XenaDataset):
         'state.treatments.diagnoses',
         'created_datetime.treatments.diagnoses',
         'updated_datetime.treatments.diagnoses',
-        'treatments.diagnoses',
         'disease_type.project',
         'exposure_id.exposures',
         'submitter_id.exposures',
--- a/xena_gdc_etl/xena_dataset.py
+++ b/xena_gdc_etl/xena_dataset.py
@@ -1289,6 +1289,7 @@ class GDCPhenoset(XenaDataset):
         'state.treatments.diagnoses',
         'created_datetime.treatments.diagnoses',
         'updated_datetime.treatments.diagnoses',
+        'disease_type.project',
         'exposure_id.exposures',
         'submitter_id.exposures',
         'state.exposures',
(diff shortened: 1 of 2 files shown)


xena_gdc_etl/xena_dataset.py
```diff
@@ -1280,6 +1280,8 @@ class GDCPhenoset(XenaDataset):
         'state.treatments.diagnoses',
         'created_datetime.treatments.diagnoses',
         'updated_datetime.treatments.diagnoses',
+        'treatments.diagnoses',
```
yunhailuo: Maybe drop this in the first place: https://github.com/ucscXena/xena-GDC-ETL/blob/master/xena_gdc_etl/gdc.py#L421 ?

maryjgoldman:
@yunhailuo I'm confused on what i'm supposed to do. It looks like the travis CI has not been run. Am I supposed to run it?

yunhailuo:
> @yunhailuo I'm confused on what i'm supposed to do. It looks like the travis CI has not been run. Am I supposed to run it?

Travis tests should run automatically if it's set up correctly. The reason I put you as reviewer for these PRs is I won't merge them until you check those results (at least a quick check) and approve them. In other words:
* I'll check the code and make sure they are right to go before requesting reviews from you.
* When you got the request for review, first check results made by this PR from Ayan;
* If the results are not good enough, we communicate and continue to change code and work on this OR;
* Once you think results are good, please approve this PR. When I saw your approval, I'll merge this PR.

Let me know if this procedure is problematic.

maryjgoldman:
That makes sense. Thank you for outlining this!

yunhailuo:
GDC has cases field on the files endpoint. They stopped embedding that info because they could make the return really big. But they haven't yet update their doc and map.

Quote on Ayan's suggestions to address:
>Then we should modify this approach https://github.com/ucscXena/xena-GDC-ETL/pull/68/files#diff-ff44a6f244700f099c71d2285046c32aR1633
>Where also we are using files endpoint with cases fields.

PR #70, opened 2019-07-03
Add XLSX file format

ayan-b:
https://github.com/ucscXena/xena-GDC-ETL/blob/58873ce30a57fd5a0a237311c419d4ba06774592/xena_gdc_etl/xena_dataset.py#L160-L161

`clinical` data now having some different sheet name other than `Clinical Data`, so I was getting errors. For example:
- `WT Data` in [Clinical.9c8108fd-c55e-46f8-ac13-9a2d9fbd1a6e.xlsx](https://github.com/ucscXena/xena-GDC-ETL/files/3354392/Clinical.9c8108fd-c55e-46f8-ac13-9a2d9fbd1a6e.xlsx) [TARGET-WT] 
- `Sample Submission` in 
[Clinical.f16dfbd6-132c-4d37-80af-357f721a1cea.xlsx](https://github.com/ucscXena/xena-GDC-ETL/files/3354413/Clinical.f16dfbd6-132c-4d37-80af-357f721a1cea.xlsx) [TARGET-AML]
- `Data Element` in 
[Clinical.74e823e2-5ffb-43ec-824c-364030dee4dc.xlsx](https://github.com/ucscXena/xena-GDC-ETL/files/3354452/Clinical.74e823e2-5ffb-43ec-824c-364030dee4dc.xlsx) [TARGET-AML] (this one is possibly not required).

### DIFF
--- a/xena_gdc_etl/xena_dataset.py
+++ b/xena_gdc_etl/xena_dataset.py
@@ -1254,16 +1254,19 @@ class GDCPhenoset(XenaDataset):
     _XENA_GDC_DTYPE = {
         'biospecimen': {
             'data_category': 'Biospecimen',
-            'data_format': 'BCR XML',
+            'data_format': ['BCR XML', 'XLSX'],
         },
-        'clinical': {'data_category': 'Clinical', 'data_format': 'BCR XML'},
+        'clinical': {
+            'data_category': 'Clinical',
+            'data_format': ['BCR XML', 'XLSX'],
+        },
         'raw_phenotype': {
             'data_category': ['Biospecimen', 'Clinical'],
-            'data_format': 'BCR XML',
+            'data_format': ['BCR XML', 'XLSX'],
         },
         'GDC_phenotype': {
             'data_category': ['Biospecimen', 'Clinical'],
-            'data_format': 'BCR XML',
+            'data_format': ['BCR XML', 'XLSX'],
         },
     }
     # To resovle overlapping between raw data and API data, remove columns
--- a/xena_gdc_etl/xena_dataset.py
+++ b/xena_gdc_etl/xena_dataset.py
@@ -163,6 +163,7 @@ def read_clinical(fileobj):
         if sheets[0] == "Clinical Data":
             return xl_file.parse(sheets[0], index_col=0)
         else:
+            print("Clincal Data not found, skipping this file ...")
             return pd.DataFrame()
     elif ext != '.xml':
         raise IOError('Unknown file type for clinical data: {}'.format(ext))
--- a/xena_gdc_etl/xena_dataset.py
+++ b/xena_gdc_etl/xena_dataset.py
@@ -158,7 +158,11 @@ def read_clinical(fileobj):
         filename = fileobj
     ext = os.path.splitext(filename)[1]
     if ext == '.xlsx':
-        return pd.read_excel(filename, sheet_name='Clinical Data', index_col=0)
+        xl_file = pd.ExcelFile(filename)
+        sheets = xl_file.sheet_names
+        if sheets[0] == "Data Elements":
+            return pd.DataFrame()
+        return xl_file.parse(sheets[0], index_col=0)
     elif ext != '.xml':
         raise IOError('Unknown file type for clinical data: {}'.format(ext))
 
@@ -1523,7 +1527,8 @@ class GDCPhenoset(XenaDataset):
             # `read_biospecimen` and `read_clinical` will check file format
             try:
                 df = read_clinical(path)
-                clin_dfs.append(df)
+                if not df.empty:
+                    clin_dfs.append(df)
             except Exception:
                 try:
                     df = read_biospecimen(path)
--- a/xena_gdc_etl/xena_dataset.py
+++ b/xena_gdc_etl/xena_dataset.py
@@ -160,9 +160,10 @@ def read_clinical(fileobj):
     if ext == '.xlsx':
         xl_file = pd.ExcelFile(filename)
         sheets = xl_file.sheet_names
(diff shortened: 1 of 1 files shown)


ayan-b:
Possible fix:
- Only consider the first list regardless of its name.
- Ignore the list if its name is `Data Elements`.

maryjgoldman:
What is this pull request doing exactly? I don't quite understand

ayan-b:
All of the raw data for TARGET clinical is in xlsx file format. So in this PR, we have added support for them.

yunhailuo:
Did a quick check on TARGET phenotype data. The file with "Data Element" is not even phenotype data. They are data description. Please hold on this and wait for more specifications. My guess is only files having "Clinical Data" sheet are real phenotype data; so I did it that way. But I need to check.

maryjgoldman:
For `Sample Submission` in Clinical.f16dfbd6-132c-4d37-80af-357f721a1cea.xlsx, let's not wrangle in this file. This file is more about QAing the sample quality rather than clinical or phenotype data our users would be interested in. Users who would be interested in this data (like someone who is developing a new algorithm to estimate tumor content), will have the skills needed to get this data directly from the GDC rather than going through us.

maryjgoldman:
@ayan-b WT Data in Clinical.9c8108fd-c55e-46f8-ac13-9a2d9fbd1a6e.xlsx [TARGET-WT] --> this looks like data on the slides. Is that true? If so, we should not wrangle it to be consistent with our wrangling of the TCGA data.

ayan-b:
Only considering `Clinical Data` now (i.e., ignoring `Data Elements` and `TARGET-WT`).

xena_gdc_etl/xena_dataset.py
```diff
@@ -158,7 +158,12 @@ def read_clinical(fileobj):
         filename = fileobj
     ext = os.path.splitext(filename)[1]
     if ext == '.xlsx':
-        return pd.read_excel(filename, sheet_name='Clinical Data', index_col=0)
+        xl_file = pd.ExcelFile(filename)
+        sheets = xl_file.sheet_names
+        if sheets[0] == "Clinical Data":
+            return xl_file.parse(sheets[0], index_col=0)
+        else:
+            return pd.DataFrame()
```
yunhailuo: If you don't want it to throw error here, remove the `try...except` below because you won't expect any, right?

xena_gdc_etl/xena_dataset.py
```diff
@@ -1520,7 +1528,8 @@ def transform(self):
             # `read_biospecimen` and `read_clinical` will check file format
             try:
                 df = read_clinical(path)
-                clin_dfs.append(df)
+                if not df.empty:
+                    clin_dfs.append(df)
```
yunhailuo: Do you still expect exception here since you checked whether it has "Clinical Data" sheet above?

yunhailuo:
Just so you know, it's probably not consistent but I code it that way to "skip" invalid phenotype xls trying to play with the [EAFP](https://docs.python.org/3/glossary.html) style.

xena_gdc_etl/xena_dataset.py
ayan-b: Removed the exception.

yunhailuo: Which exception did you remove? You checked the sheet name in `read_clinical` so that there is no exception yet you remove `try...except` clause for `read_biospecimen`? Can you elaborate the logic here?

xena_gdc_etl/xena_dataset.py
yunhailuo: Also, the file is essentially skipped and not used/processed here. Since it won't raise any exceptions, it's better to leave a message here showing you "skip" this file.

xena_gdc_etl/xena_dataset.py
ayan-b: OK, I had misunderstood.
> You checked the sheet name in read_clinical so that there is no exception 

Agreed, `if not df.empty` is redundant and can be removed ...

yunhailuo: Sorry. My fault. You should probably put it back to 4724e4cbe3037cffe3b1ac8dc983c01cd4edc566. Will try to explain myself below.

yunhailuo:
OK. I didn't make this clear. Let me try this. This is a design problem. Let's imagine clinical data is a dict for simplification.

For [EAFP style](https://docs.python.org/3/glossary.html#term-eafp), you don't check whether you have "Clinical Data" or not but just get it and catch error later:

```python
def read_clinical(fileobj):
    return fileobj['Clinical Data']


dummy_fileobj = {
    'No Clinical Data': 'Nothing'
}
clin_dfs = []
try:
    df = read_clinical(dummy_fileobj)
    clin_dfs.append(df)
except Exception:
    print('No clinical data in this file. Skip and move on')
```

For [LBYL style](https://docs.python.org/3/glossary.html#term-lbyl), you check before doing anything and you don't have to catch error later because there is no error expected:

```python
def read_clinical(fileobj):
    if 'Clinical Data' not in fileobj:
        return
    return fileobj['Clinical Data']


dummy_fileobj = {
    'No Clinical Data': 'Nothing'
}
clin_dfs = []
# The following line will work no matter what keys are in dummy_fileobj
df = read_clinical(dummy_fileobj)
if df:
    clin_dfs.append(df)
else:
    print('No clinical data in this file. Skip and move on')
```

I thought you are trying to change EAFP to LBYL in this PR. But I was wrong. My initial implementation has flaws and your 3c0f539c5255c7e357c5e60c08b0cfacbacc200c should be right.

PR #82, opened 2019-07-11
Remove columns having `list` datatype

### LINKED ISSUE #64: Remove disease_type.project from all phenotype files
I think we said to do this at some point but I'm still seeing this field in the files. It is a list and we don't want that. And the data in this field is already contained in the 'disease_type' field that we get from the xml files.

ayan-b:
Closes #64.

### DIFF
--- a/xena_gdc_etl/gdc.py
+++ b/xena_gdc_etl/gdc.py
@@ -418,7 +418,6 @@ def get_samples_clinical(projects=None):
     expand = [
         'demographic',
         'diagnoses',
-        'diagnoses.treatments',
         'exposures',
         'family_histories',
         'project',
--- a/xena_gdc_etl/gdc.py
+++ b/xena_gdc_etl/gdc.py
@@ -427,10 +427,10 @@ def get_samples_clinical(projects=None):
     res = search(
         'cases', in_filter=in_filter, fields=fields, expand=expand, typ='json'
     )
-    reduced_json = reduce_json_array(res)
     to_drops = set()
-    for ele in reduced_json:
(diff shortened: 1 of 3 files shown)


yunhailuo:
Commented in slack. Remove it before making dataframe: https://github.com/ucscXena/xena-GDC-ETL/blob/master/xena_gdc_etl/gdc.py#L431-L434. Checking for "[" is hacky and not right.

xena_gdc_etl/gdc.py
```diff
@@ -447,7 +450,9 @@ def get_samples_clinical(projects=None):
         'id',
         record_prefix='samples.',
     )
-    return pd.merge(cases_df, samples_df, how='inner', on='id')
+    merged_df = pd.merge(cases_df, samples_df, how='inner', on='id')
+    merged_df.drop(list(to_drops), axis=1, inplace=True)
```
yunhailuo: Those fields can be dropped in JSON and don't have to wait until here. Dropping something as early as possible would save extra computation.

xena_gdc_etl/xena_dataset.py
```diff
@@ -28,7 +28,11 @@
 import pandas as pd
 
 from xena_gdc_etl import gdc
-from .utils import mkdir_p, requests_retry_session, reduce_json_array
+from .utils import (
+    mkdir_p,
+    requests_retry_session,
+    reduce_json_array,
+)
```
yunhailuo: Is this change related to the features here? I'd prefer no unrelated changes in PR and code review.

PR #83, opened 2019-07-11
Fix non-concatenation axis non-aligned warning

ayan-b:
Fix this `FutureWarning`:

```
FutureWarning: Sorting because non-concatenation axis is not aligned. A future version
of pandas will change to not sort by default.

To accept the future behavior, pass 'sort=False'.

To retain the current behavior and silence the warning, pass 'sort=True'.

  pd.concat(bio_dfs, axis=0)
```


yunhailuo:
Ready? Done for all? We have more than two `pd.concat`, right?

yunhailuo:
For each case, is `sort=True` really needed? And which one will save resources, `True` or `False`?

ayan-b:
> Ready? Done for all? We have more than two pd.concat, right?

I don't think I have seen errors from them.

ayan-b:
> For each case, is sort=True really needed? And which one will save resources, True or False?

I guess `False`.

yunhailuo:
> > Ready? Done for all? We have more than two pd.concat, right?
> 
> I don't think I have seen errors from them.

This is not an error. It's a deprecation notice. If we really need sort here, we pass `sort=True` and if we don't, we don't have to do anything regarding to this warning. So
* Do we really need `sort=True` here?
* Do we need `sort=True` for other `pd.concat`?

Let do a fix really fix the root problem rather than simply making a warning disappear.

ayan-b:
Yes, makes sense keep this as it is.

PR #84, opened 2019-07-13
Shorten codes for list removal on GDC API phenotype data

yunhailuo:


yunhailuo:
TCGA-CHOL phenotype matrix is exactly the same. Will probably check a few more cohorts before merging.

PR #85, opened 2019-07-15
Drop samples having no genomic data

### LINKED ISSUE #63: Samples with phenotype data but no genomic data are still being included

ayan-b:
Closes #63.

### DIFF
--- a/xena_gdc_etl/xena_dataset.py
+++ b/xena_gdc_etl/xena_dataset.py
@@ -1704,6 +1704,8 @@ class GDCPhenoset(XenaDataset):
                     'Getting "GDC_phenotype" for a cohort with mixed TCGA and '
                     'TARGET projects is not currently suppported.'
                 )
+        print("Dropping samples having no genomic data ...")
+        xena_matrix = xena_matrix[~xena_matrix.index.str.endswith('Z')]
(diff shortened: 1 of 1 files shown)


xena_gdc_etl/xena_dataset.py
```diff
@@ -1704,6 +1704,8 @@ def transform(self):
                     'Getting "GDC_phenotype" for a cohort with mixed TCGA and '
                     'TARGET projects is not currently suppported.'
                 )
+        print("Dropping samples having no genomic data ...")
```
yunhailuo: Not essential. But it would be better if the message describes what really happens here. It's not actually targeting specifically all samples with no genomic data...

ayan-b: Updated the message.

PR #90, opened 2019-07-23
Reorder merge

ayan-b:


xena_gdc_etl/xena_dataset.py
```diff
@@ -1631,23 +1638,6 @@ def transform(self):
                 xena_matrix = bio_matrix.set_index('bcr_sample_barcode')
             except Exception:
                 xena_matrix = bio_matrix
-        elif self.xena_dtype in ['raw_phenotype', 'GDC_phenotype'] and all(
```
ayan-b: Wondering what to do with `raw_phenotype` now ...

yunhailuo: This is a bad design from me at the very beginning. Sorry. That's why I want to refactor the code. Right now, a slightly cleaner way is to separate `GDC_phenotype` (with TCGA if needed) out in another elif condition. I have to say I don't know if that's clean enough so don't take it as a sure thing. Try and let us know.

Overall, the feature being developed in this PR is
* Targeting only GDC_phenotype for all and just TCGA projects
* Output GDC_phenotype matrices should be the same for all TCGA projects other than BRCA, meaning needing some tests (not writing test code but doing a few test runs and compare results)
* Trying not to affect the logic on GDC_phenotype for all TARGET projects
* Not caring about others, like `raw_phenotype`, `clinical`, etc. Those datatype are not real or used by our work. So no rule for them. We can change their behavior as we want or even remove them (probably cleaner this way?).

xena_gdc_etl/xena_dataset.py
```diff
@@ -1631,23 +1638,6 @@ def transform(self):
                 xena_matrix = bio_matrix.set_index('bcr_sample_barcode')
```
yunhailuo: Will probably break here due to the change above. But it's fine to leave it like this for now. This is due to the terrible design I had at the very beginning.

xena_gdc_etl/xena_dataset.py
ayan-b: e9169f2 probably fixes this `raw_phenotype` issue.

PR #91, opened 2019-07-30
Add GDCAPIPhenoset

### LINKED ISSUE #57: Add CPTAC cohort
Should be just RNAseq data and clinical data from the API (no xml files)

ayan-b:
- [x] Add a note mentioning why some UUIDs are in `samples.submitter_id`. Examples: a seemingly improper case: https://portal.gdc.cancer.gov/cases/ba0fe300-d3f0-42fd-b609-3ee3d0e49d7b and a proper one: https://portal.gdc.cancer.gov/cases/4c241b93-c11c-4802-94a7-07f125267ba3.

Closes #57.


ayan-b:
Once the code part is finalized, I will add the docstrings.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,162 @@ def transform(self):
...
+            return self.__metadata_vars
+
+    @metadata_vars.setter
+    def metadata_vars(self, variables):
+        self.__metadata_vars = variables
+
+    @XenaDataset.download_map.getter
+    def download_map(self):
+        print("Xena_phenotype is selected. No files will be downloaded.")
+        return {}
```
yunhailuo: The message here is very nice and considerable.

My logic here is user may still set and use the `download_map` **property**. Banning it so that user can only ever get `{}` from it is OK. I just have a little concern about the behavior. If I set `download_map` to a dict I like (the setter seems still working) and end up seeing an empty dict why trying to get it, I would start to wonder why my setting doesn't work. Maybe either allow users to play with `download_map` whatever they like since it hurts nothing, does nothing and confuse nobody (might be confusing a little bit because it seems useless...) since it's a property or ban it completely. I'm not sure which one it's better but I think the first option requires no code.

I'm also wondering (or more worrying) about the download method. Users can still use that method without error. It actually does nothing real but set `self.raw_data_list` and print out misleading messages suggesting something have been done: https://github.com/ucscXena/xena-GDC-ETL/blob/master/xena_gdc_etl/xena_dataset.py#L677-L707

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,162 @@ def transform(self):
...
+                fields=CASES_FIELDS_EXPANDS["CPTAC-3"]["fields"],
+                expand=CASES_FIELDS_EXPANDS["CPTAC-3"]["expand"],
+            )
+            xena_matrix = xena_matrix.set_index('samples.submitter_id')
+        else:
+            raise NotImplementedError(
+                "Xena_phenotype for {} project(s) is not "
+                "implemented".format(",".join(self.projects))
+            )
+        print('\rSaving matrix to {} ...'.format(self.matrix), end='')
```
yunhailuo: A note only: I reversed the hierarchy order in column names before because I felt it's annoying to have the real important name at the end of the string (especially when some strings are long) back then. But I do start to feel it's unnecessary or even wrong now. So let do it here without reversing and see how Mary likes it.

ayan-b:
The main purpose of this PR is to get the phenotype data via the API only (no downloading of XML files). If we know that if API data from a project (TCGA, TARGET) is not sufficient, we will not allow it here.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,162 @@ def transform(self):
...
+
+    def __init__(
+        self,
+        projects,
+        xena_dtype=None,
+        root_dir='.',
+        raw_data_dir=None,
+        matrix_dir=None,
+    ):
+        self.projects = projects
```
yunhailuo: Check projects here.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,162 @@ def transform(self):
...
+        if xena_dtype is not None:
+            if xena_dtype != 'Xena_phenotype':
+                raise ValueError(
+                    "No datatype other than 'Xena_phenotype' is supported."
+                )
+            else:
+                self.xena_dtype = xena_dtype
```
yunhailuo: Should ban this input if only one xena_dtype is allowed for this class.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,162 @@ def transform(self):
...
+            self.xena_dtype = 'Xena_phenotype'
+        self.root_dir = root_dir
+        if matrix_dir is not None:
+            self.matrix_dir = matrix_dir
+        jinja2_env = jinja2.Environment(
+            loader=jinja2.PackageLoader('xena_gdc_etl', 'resources')
+        )
+        self.metadata_template = jinja2_env.get_template(
+            'template.phenotype.meta.json'
+        )
```
yunhailuo: If I got this right, most of this code should have been done through `super().__init__()`.

xena_gdc_etl/xena_dataset.py
yunhailuo: Based on the PR spec, `download` and all related things should for sure be banned since it's API only.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,162 @@ def transform(self):
...
+        merged_df = pd.merge(cases_df, samples_df, how='inner', on='id')
+        merged_df.drop(list(to_drops), axis=1, inplace=True)
+        return merged_df
+
+    def __init__(
+        self,
+        projects,
+        xena_dtype=None,
+        root_dir='.',
+        raw_data_dir=None,
```
yunhailuo: No raw data for this?

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,162 @@ def transform(self):
...
+            self.matrix_dir = matrix_dir
+        jinja2_env = jinja2.Environment(
+            loader=jinja2.PackageLoader('xena_gdc_etl', 'resources')
+        )
+        self.metadata_template = jinja2_env.get_template(
+            'template.phenotype.meta.json'
+        )
+
+    def transform(self):
+        if len(self.projects) == 1 and self.projects[0] == "CPTAC-3":
```
yunhailuo: Why not directly `self.projects == 'CPTAC-3'`?

Moreover, why if here since `CASES_FIELDS_EXPANDS` looks like your switch for different projects with this class focus on API phenotype?

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,162 @@ def transform(self):
...
+                projects=["CPTAC-3"],
+                fields=CASES_FIELDS_EXPANDS["CPTAC-3"]["fields"],
+                expand=CASES_FIELDS_EXPANDS["CPTAC-3"]["expand"],
+            )
+            xena_matrix = xena_matrix.set_index('samples.submitter_id')
+        else:
+            raise NotImplementedError(
+                "Xena_phenotype for {} project(s) is not "
+                "implemented".format(",".join(self.projects))
+            )
```
yunhailuo: Maybe this should be done in `init` so that resources won't be wasted for a doomed failure until here.

xena_gdc_etl/xena_dataset.py
ayan-b: Banning the input now.

xena_gdc_etl/xena_dataset.py
ayan-b: Checking projects here now.

xena_gdc_etl/xena_dataset.py
ayan-b: Removed the `raw_data` option.

xena_gdc_etl/xena_dataset.py
ayan-b: Doing it in `init` now.

xena_gdc_etl/xena_dataset.py
ayan-b: Not reversing the column names.

xena_gdc_etl/xena_dataset.py
ayan-b: Yes, its is banned.

xena_gdc_etl/gdc2xena.py
```diff
@@ -87,6 +92,8 @@ def gdc2xena(root_dir, projects, xena_dtypes, delete_raw_data=False):
                     dataset = GDCPhenoset(project, 'clinical', root_dir)
             elif dtype == 'GDC_phenotype':
                 dataset = GDCPhenoset(project, 'GDC_phenotype', root_dir)
+            elif dtype == 'Xena_phenotype':
+                dataset = XenaPhenoset(project, 'Xena_phenotype', root_dir)
```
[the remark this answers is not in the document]
ayan-b: Wondering if we should use `GDC_phenotype` for the API only data ...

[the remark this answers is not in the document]
yunhailuo: We can switch to Xena_phenotype. Just make it consistent.

Another thing is XenaPhenoset is misleading since this class/PR is designed to deal with API only rather than accommodating all XenoPhenoset cases.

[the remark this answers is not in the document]
ayan-b: I guess it can be`GDCAPIPhenoset`, and should we change the datatype to GDC_API_phenotype ... probably too long.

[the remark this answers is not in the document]
yunhailuo: I won't suggest that. `GDCAPIPhenoset` is an internal thing for implementation which is for code reader who cares about implementation. It should have details and be explicit. The datatype, as mentioned above, is for users who should not care about details of a good pipeline implementation. I would suggest it to be simple, easy to understand and stable.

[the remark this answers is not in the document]
ayan-b: OK, using `Xena_phenotype` and renamed `XenaPhenoset` to `GDCAPIPhenoset`.

xena_gdc_etl/xena_dataset.py
ayan-b: So, we should enforce user to choose one project only?

For example, we can do:
```python
xena_matrix = self.__get_samples_clinical(
                projects=self.projects[0],
                fields=CASES_FIELDS_EXPANDS[self.projects[0]]["fields"],
                expand=CASES_FIELDS_EXPANDS[self.projects[0]]["expand"],
            )
```
If we need to do some operations on certain matrices we can do them in an `if` clause later in the code.

yunhailuo: My understanding is this class suppose to handle any projects whose phenotype is GDC API only (even though they may require different fields, which is actually configured by `CASES_FIELDS_EXPANDS`). So it's not "one project only" but all projects, if more than one should be **valid** (need GDC API only). Whether CPTAC should be combinable with other projects (not really but say TCGA as an example) is user's decision because those pandas dataframes are technically combinable as far as the coding concerns.

ayan-b: > Moreover, why if here since `CASES_FIELDS_EXPANDS` looks like your switch for different projects with this class focus on API phenotype?

I am not able to understand this line. :worried:

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,173 @@ def transform(self):
...
+            fields=fields,
+            expand=expand,
+            typ='json'
+        )
+        to_drops = set()
+        for ele in res:
+            to_drops |= set(gdc.get_to_drops(ele))
+        print(
+            "Dropping columns {} for {} projects".format(to_drops, projects)
+        )
```
yunhailuo: I'm prone to use explicit field list and 1) not to use expand; 2) not to drop columns (fields) after API requests. Phenotype data is complicated and knowing what to do before doing things seems like a good practice here. We can leave it like this for now though. I can clean it up later.

xena_gdc_etl/xena_dataset.py
ayan-b: Do you mean:
- Build a single dataframe for a single project and build those one by one.
- Merge those dataframes into a final dataframe.
- Return the dataframe after some operations (if needed).

yunhailuo: Valid projects are keys of `CASES_FIELDS_EXPANDS` if I understand it correctly. So hard code "CPTAC-3" here seems a bad practice.

Well, it's like how data are merged for TCGA XML dataframes. But let's restrain it to CPTAC-3  here and I'll fix that later.

yunhailuo: So no changes but just move it to `init()` for now.

xena_gdc_etl/xena_dataset.py
```diff
@@ -1719,6 +1720,173 @@ def transform(self):
...
+            raise NotImplementedError(
+                "'Xena_phenotype' for {} project is not implemented".format(
+                    projects
+                )
+            )
+        jinja2_env = jinja2.Environment(
+            loader=jinja2.PackageLoader('xena_gdc_etl', 'resources')
+        )
+        self.metadata_template = jinja2_env.get_template(
+            'template.Xena_phenotype.meta.json'
```
yunhailuo: This is the template name you want for `GDCAPIPhenoset`?

yunhailuo: And why it needs to be different from the other phenotype template?

ayan-b: There is an extra `description` key.

yunhailuo: And is that specific for CPTAC-3 or is that general for any GDC API phenotypes or Xena phenotypes? I'm wondering what is the correct info matching your abstraction level here, meaning this PR and class is designed to be general rather than CPTAC-3 only, right?

ayan-b: Sorry. I should have made that for `CPTAC-3` only. Thanks.

ayan-b: Wondering what is the best way to do that? I can pass `project` and check that, for example:
```{% if project == "CPTAC-3"%}
description: ...
{%endif%}
```.

Comment (yunhailuo) [reply xena_gdc_etl/xena_dataset.py]: Just be sure about your design and overall logic. I would have approved your PR weigh before if we just want some code working. But I guess that's not our final goal here.

File: xena_gdc_etl/xena_dataset.py (hunk 7 above)
Comment (ayan-b) [reply xena_gdc_etl/xena_dataset.py]: There is a check in the `init()`:
```python
        if any(
            [
                project not in list(CASES_FIELDS_EXPANDS.keys())
                for project in self.projects
            ]
        ):
            raise NotImplementedError(
                "'Xena_phenotype' for {} project is not implemented".format(
                    projects
                )
            )
```

yunhailuo: OK. Good. Then why you would need a check here?

xena_gdc_etl/xena_dataset.py
yunhailuo: It looks like a project specific metadata_var to me.

xena_gdc_etl/xena_dataset.py
ayan-b: In future, if we include other projects (say `MMRF-COMMPASS`) and then we can add them in another if clause.
```python
if self.projects == ["CPTAC-3"]:
   # make matrix
   ...
elif self.projects == ["MMRF-COMMPASS"]:
  # make matrix
  ...
```

yunhailuo: Why would they be handled differently if they both use only GDC API and have fields defined clearly in `CASES_FIELDS_EXPANDS`?

ayan-b: You can mean it can be handled like this: https://github.com/ucscXena/xena-GDC-ETL/pull/91#discussion_r310351533
Also,
> If we need to do some operations on certain matrices we can do them in an `if` clause later in the code.

by later I mean after generating the `xena_matrix`.

yunhailuo: I couldn't see any use case for this code. Like `self.projects[0]` is unnecessarily specific and I don't know what you would do other than saving it after making your matrix. In general, I wouldn't recommend any code without clear usage.

But it works anyway, so just leave it there and we can move on.

PR #95, opened 2019-08-07
Drop list fields from XML files

ayan-b:


ayan-b:
Just realized a it'd be good idea to check in the read_clinical function itself.

PR #102, opened 2019-09-04
Improve GDCAPIPhenoset

ayan-b:
Realized that I coded this up but never made a PR!

In this PR:
- Hard-coded projects are removed
- Multiple projects in GDCAPIPhenoset are now supported
- Removed expand from GDC-PANCAN.


ayan-b:
Here is a list of fields for those new projects which can be incorporated later.

[fields.txt](https://github.com/ucscXena/xena-GDC-ETL/files/3572683/fields.txt)

